PR #19, opened 2015-04-30
upipe_rtp_prepend: support timestamp sync on pts

quarium:


cmassiot:
Probably we'd better use the flow definition to determine the stream_type, the clock frequency and the clock source.

PR #23, opened 2015-05-04
add upipe_rtp_h264

quarium:


cmassiot:
Why not use the h264 framer before this pipe ? It would avoid the ugly uref_block_extract and start code search (that are already duplicated there btw).

PR #34, opened 2015-06-01
add a pipe to dump input uref

quarium:


cmassiot:
Your urequest handling is not correct : you should register the incoming urequest to your output.

quarium:
Is it correct now ?

PR #201, opened 2016-06-17
upipe-av: decode subtitles

funman:


include/upipe/ubuf_pic.h
```diff
@@ -388,7 +397,69 @@ static inline int ubuf_pic_blit(struct ubuf *dest, struct ubuf *src,
         int plane_vsize = extract_vsize / src_vsub;
 
         for (int i = 0; i < plane_vsize; i++) {
-            memcpy(dest_buffer, src_buffer, plane_hsize);
+            if (!alpha_plane || threshold == 0) {
+                memcpy(dest_buffer, src_buffer, plane_hsize);
+            } else if (threshold != 0xff) {
+                for (int j = 0; j < plane_hsize; j += 16) {
```
funman: How do we force stride alignement to avoid a final loop with less than 16 pixels?

cmassiot: you can ask for an aligned buffer in upipe_blit with a urequest flow format, but I don't think ubuf_pic_blit should rely on it.

cmassiot: handle subpicture expiration: I'd prefer to keep the old behaviour (subpics are blitted until they are replaced) because some people rely on it (upipe_blit is used for other things than subtitles). You could instead have a pipe in front of upipe_blit that checks duration and emits an empty uref when the subtitle is over.

include/upipe/ubuf_pic.h
```diff
@@ -388,7 +397,69 @@ static inline int ubuf_pic_blit(struct ubuf *dest, struct ubuf *src,
...
+                    /* This is an on/off blending
+                     * if alpha is over the threshold, we use the subpicture pixel.
+                     */
+                    const uint8_t *a = &alpha_plane[alpha_stride * (i * src_vsub) + j * src_hsub];
+                    if (a[0x0] > threshold) dest_buffer[j+0x0] = src_buffer[j+0x0];
```
cmassiot: any reason not to use a for() loop here ?

lib/upipe-modules/upipe_blit.c
```diff
@@ -218,6 +246,22 @@ static void upipe_blit_sub_input(struct upipe *upipe, struct uref *uref,
...
+        duration = 0;
+
+	uref_clock_get_date_prog(uref, &sub->date_prog, &type);
+	if (type == UREF_DATE_NONE) {
+        sub->date_prog = UINT64_MAX;
+        sub->end_date = UINT64_MAX;
+    } else {
+        sub->end_date = sub->date_prog + duration;
+    }
+
```
cmassiot: I'd prefer to keep the old behaviour (subpics are blitted until they are replaced) because some people rely on it (upipe_blit is used for other things than subtitles). You could instead have a pipe in front of upipe_blit that checks duration and emits an empty uref when the subtitle is over. BTW, in upipe_blit there is no check of the input PTS to compare it with the subpic PTS. So to have precise subtitling you need to rework your architecture. We can discuss that on IRC.

lib/upipe-av/upipe_avcodec_encode.c
```diff
@@ -509,31 +509,35 @@ static bool upipe_avcenc_encode_frame(struct upipe *upipe,
...
+        av_packet_unref(&avpkt);
         upipe_throw_fatal(upipe, UBASE_ERR_ALLOC);
         return false;
     }
     memcpy(buf, avpkt.data, size);
     ubuf_block_unmap(ubuf, 0);
-    av_free_packet(&avpkt);
+
+    int64_t pkt_pts = avpkt.pts, pkt_dts = avpkt.dts;
+    bool keyframe = avpkt.flags & AV_PKT_FLAG_KEY;
```
cmassiot: I think you are supposed to use !!() to convert to boolean.

include/upipe/ubuf_pic.h
funman: benchmarking shows that at -O2 unrolling 4 loops is 4 times faster, unrolling more loops has less of an effect and can even be worse at -O3.

I will add tests to that commit log.

Also unrolling only helps for the threshold method, it doesn't do much for the division method, so it can be removed for that one

include/upipe/ubuf_pic.h
funman: True, I can change it into unrolling 4 loops, plus a termination loop for the last pixels

lib/upipe-av/upipe_avcodec_encode.c
funman: AFAIK !! makes sure the boolean is 0 or 1 rather than something > 0 that still makes the boolean true, so it should be good if we don't write that boolean as 32 bits into a bitstream

PR #207, opened 2016-07-18
Bring back uref_block_extend from 55001e8

funman:
This can be used instead of less performant iovec when writing ubufs to DMA

### DIFF
--- a/include/upipe/ubuf_block.h
+++ b/include/upipe/ubuf_block.h
@@ -479,6 +479,34 @@ static inline int ubuf_block_resize(struct ubuf *ubuf, int offset, int new_size)
     return UBASE_ERR_NONE;
 }
 
+/** @This prepends a block ubuf, if possible. This will only work if
+ * prepend has been correctly specified at allocation.
+ *
+ * Should this fail, @ref ubuf_block_merge may be used to achieve the same
+ * goal with an extra buffer copy.
+ *
+ * @param ubuf pointer to ubuf
+ * @param prepend number of octets to prepend
+ * @return an error code
+ */
+static inline int ubuf_block_prepend(struct ubuf *ubuf, int prepend)
+{
+    assert(prepend >= 0);
+    if (ubuf->mgr->signature != UBUF_ALLOC_BLOCK)
+        return UBASE_ERR_INVALID;
+
+    struct ubuf_block *block = ubuf_block_from_ubuf(ubuf);
+
+    if (prepend > block->offset)
+        return UBASE_ERR_INVALID;
+
+    block->offset -= prepend;
+    block->size += prepend;
+    block->total_size += prepend;
+    block->cached_offset += prepend;
+    return UBASE_ERR_NONE;
+}
+
 /** @This duplicates part of a ubuf.
  *
  * @param ubuf pointer to ubuf
--- a/lib/upipe/ubuf_block_mem.c
+++ b/lib/upipe/ubuf_block_mem.c
@@ -49,6 +49,8 @@
 
 /** default alignement of buffer when unspecified */
 #define UBUF_DEFAULT_ALIGN          0
+/** default minimum extra space before buffer when unspecified */
+#define UBUF_DEFAULT_PREPEND        32
 
 /** @This is a super-set of the @ref ubuf (and @ref ubuf_block)
  * structure with private fields pointing to shared data. */
@@ -68,6 +70,8 @@ struct ubuf_block_mem_mgr {
     /** refcount management structure */
     struct urefcount urefcount;
 
+    /** extra space added before */
+    size_t prepend;
     /** alignment */
     size_t align;
     /** alignment offset */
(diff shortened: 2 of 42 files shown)


lib/upipe/ubuf_mem.c
```diff
@@ -64,7 +64,7 @@ struct ubuf_mgr *ubuf_mem_mgr_alloc_from_flow_def(uint16_t ubuf_pool_depth,
         uref_block_flow_get_align(flow_def, &align);
         uref_block_flow_get_align_offset(flow_def, &align_offset);
         return ubuf_block_mem_mgr_alloc(ubuf_pool_depth, shared_pool_depth,
-                                        umem_mgr, align, align_offset);
+                                        umem_mgr, 0, 0, align, align_offset);
```
cmassiot: You will need to replace 0, 0 with values retrieved from flow_def attributes.

include/upipe/ubuf_block.h
```diff
@@ -479,6 +479,50 @@ static inline int ubuf_block_resize(struct ubuf *ubuf, int offset, int new_size)
...
+ * @return an error code
+ */
+static inline int ubuf_block_extend(struct ubuf *ubuf, int prepend, int append)
+{
+    assert(prepend >= 0);
+    assert(append >= 0);
+    if (ubuf->mgr->signature != UBUF_ALLOC_BLOCK)
+        return UBASE_ERR_INVALID;
+
+    if (!ubase_check(ubuf_control(ubuf, UBUF_SINGLE)))
```
cmassiot: I don't think you need to check that the block is not duplicated at this point. It will be done when you call ubuf_block_write() anyway.

include/upipe/ubuf_block.h
```diff
@@ -479,6 +479,50 @@ static inline int ubuf_block_resize(struct ubuf *ubuf, int offset, int new_size)
...
+        return UBASE_ERR_BUSY;
+
+    struct ubuf_block *block = ubuf_block_from_ubuf(ubuf);
+    if (prepend > block->offset)
+        return UBASE_ERR_INVALID;
+    if (append) {
```
cmassiot: I am not sure we need to get into that logic. Just see if there is enough contiguous space to expend the block, or bail out and let the caller call ubuf_block_merge. This would also get rid of the additional routine in ubuf_block_mem.c.

lib/upipe/ubuf_block_mem.c
```diff
@@ -216,6 +225,38 @@ static int ubuf_block_mem_splice(struct ubuf *ubuf, struct ubuf **new_ubuf_p,
...
+}
+
+/** @internal @This extends a block_mem ubuf.
+ *
+ * @param ubuf pointer to ubuf
+ * @param new_size required size of the buffer
+ * @return false in case of error, or if the ubuf is shared, or if the operation
+ * is not possible
+ */
+static bool ubuf_block_mem_extend(struct ubuf *ubuf, int new_size)
```
cmassiot: As said earlier, I don't think we need to bother with this case.

lib/upipe/ubuf_block_mem.c
```diff
@@ -216,6 +225,38 @@ static int ubuf_block_mem_splice(struct ubuf *ubuf, struct ubuf **new_ubuf_p,
...
+{
+    if (!ubuf_block_mem_single(ubuf))
+        return false;
```
cmassiot: Be careful, the API has changed, it is now UBASE_ERR_BUSY. See ubuf_pic_resize, which is the most similar function to what you are trying to achieve.

include/upipe/ubuf_block.h
funman: I don't understand, ubuf_block_write is not called in this function.
We wouldn't want to change an ubuf used by another uref.

That UBUF_SINGLE check was added in 5b7253903251

lib/upipe/ubuf_block_mem.c
funman: Yep, big mistake on my part

lib/upipe/ubuf_mem.c
funman: Ok I need to also define the attributes in ubuf_mem.h

include/upipe/ubuf_block.h
funman: Do you mean to refuse appending to segmented blocks?

include/upipe/ubuf_block.h
cmassiot: I suppose the use case is, have a payload, extend it upwards, call ubuf_block_write to write some header information (or footer). So there will be a ubuf_block_write, and it's the one that needs to check UBUF_SINGLE. Your function has no influence on the refcounted buffer.

include/upipe/ubuf_block.h
cmassiot: I don't see a use case for it. And it adds a lot of complexity and a ubuf control command. Perhaps it can be resurrected later if needs be.

lib/upipe/ubuf_mem.c
cmassiot: No in uref_block_flow.h

include/upipe/ubuf_block.h
funman: OK that makes sense.
However I don't see if I can find the allocated buffer size (versus used buffer size) to check if I can append to the block.
Did I miss something or do I need to add the allocated size to struct ubuf_block?

cmassiot: That's correct, there is no way to know. So actually what you did was the best way to do it.

PR #208, opened 2016-07-19
Luajit bindings

nto:


luajit/uprobe-args.lua
```diff
@@ -0,0 +1,20 @@
+return {
+    -- upipe
```
cmassiot: Perhaps we should be a way to auto generate this from C headers ?

PR #209, opened 2016-08-18
rtp_prepend: fix mpa payload

funman:
https://tools.ietf.org/html/rfc2250#section-3.5

### DIFF
--- a/lib/upipe-modules/upipe_rtp_prepend.c
+++ b/lib/upipe-modules/upipe_rtp_prepend.c
@@ -167,6 +167,33 @@ static void upipe_rtp_prepend_input(struct upipe *upipe, struct uref *uref,
     ubuf_block_unmap(header, 0);
     upipe_rtp_prepend->seqnum++;
 
+    const char *def;
+    ubase_assert(uref_flow_get_def(upipe_rtp_prepend->flow_def, &def));
+    if (!ubase_ncmp(def, "block.rtp.mp2.sound.") ||
+            !ubase_ncmp(def, "block.rtp.mp3.sound.")) {
+        /* alloc mpa header */
+        struct ubuf *mpa_header = ubuf_block_alloc(uref->ubuf->mgr, 4);
+        if (unlikely(!mpa_header)) {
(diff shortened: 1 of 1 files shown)


lib/upipe-modules/upipe_rtp_prepend.c
```diff
@@ -167,6 +167,33 @@ static void upipe_rtp_prepend_input(struct upipe *upipe, struct uref *uref,
     ubuf_block_unmap(header, 0);
     upipe_rtp_prepend->seqnum++;
 
+    const char *def;
+    ubase_assert(uref_flow_get_def(upipe_rtp_prepend->flow_def, &def));
+    if (!ubase_ncmp(def, "block.rtp.mp2.sound.") ||
+            !ubase_ncmp(def, "block.rtp.mp3.sound.")) {
+        /* alloc mpa header */
+        struct ubuf *mpa_header = ubuf_block_alloc(uref->ubuf->mgr, 4);
```
cmassiot: Unfortunately, you are not allowed to do that :(. The reason is that ubuf->mgr could be a ubuf manager from a hardware demux, with buffer backed by video or special memory, so it may not be able to allocate anything, and it may not be efficient.
The correct way is the more cumbersome urequest with a ubuf manager that we do everywhere else.

lib/upipe-modules/upipe_rtp_prepend.c
```diff
@@ -167,6 +167,33 @@ static void upipe_rtp_prepend_input(struct upipe *upipe, struct uref *uref,
...
+        if (unlikely(!mpa_header)) {
+            upipe_throw_fatal(upipe, UBASE_ERR_ALLOC);
+            uref_free(uref);
+            ubuf_free(header);
+            return;
+        }
+
+        /* write mpa mpa_header */
+        ubuf_block_write(mpa_header, 0, &size, &buf);
+        memset(buf, 0, 4); // frag_offset = 0
```
cmassiot: Here, it may be more efficient to allocate the ubuf just once, and for every subsequent packet call ubuf_dup() to get a shared pointer to the same area. That's what we do for TS padding for instance.

PR #213, opened 2016-08-24
Bmdsrc leak

funman:
fix memory leaks detected with valgrind --undef-value-errors=no --leak-check=full

### DIFF
--- a/lib/upipe-blackmagic/upipe_blackmagic_source.cpp
+++ b/lib/upipe-blackmagic/upipe_blackmagic_source.cpp
@@ -1146,6 +1146,9 @@ static void upipe_bmd_src_free(struct upipe *upipe)
         upipe_bmd_src->deckLink->Release();
     uqueue_clean(&upipe_bmd_src->uqueue);
 
+    ubuf_mgr_release(upipe_bmd_src->pic_subpipe.ubuf_mgr);
+    ubuf_mgr_release(upipe_bmd_src->sound_subpipe.ubuf_mgr);
+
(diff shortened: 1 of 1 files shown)


lib/upipe-blackmagic/upipe_blackmagic_source.cpp
```diff
@@ -821,8 +821,10 @@ static int upipe_bmd_src_set_uri(struct upipe *upipe, const char *uri)
     }
 
     const char *model_name;
-    if (deckLink->GetModelName(&model_name) == S_OK)
+    if (deckLink->GetModelName(&model_name) == S_OK) {
         upipe_notice_va(upipe, "detected card type %s", model_name);
+        free((void*)model_name);
```
cmassiot: Then it's not a const char * ? The SDK doc says string *, any idea what this represents ? Otherwise change const char * to char * to avoid the cast. Perhaps we should use delete instead of free as well.

funman: It's declared as const char ** in the header:
virtual HRESULT GetModelName (/* out */ const char **modelName) = 0;

Here is some code copy/pasted from blackmagic own example Capture:

        deckLink->GetModelName((const char**)&m_deckLinkName);
        free(m_deckLinkName);
        char* m_deckLinkName;

As you see they cast the other way around...

cmassiot: Then do as they do (crazy people).

PR #217, opened 2016-10-05
Udp

funman:
Add more verbose rtp/udp debug

Fix a couple problems when using ipv6 sockets

### DIFF
--- a/lib/upipe-modules/upipe_udp.c
+++ b/lib/upipe-modules/upipe_udp.c
@@ -391,7 +391,8 @@ int upipe_udp_open_socket(struct upipe *upipe, const char *_uri, int ttl,
             return -1;
         }
         /* required on some architectures */
-        memset(&connect_addr.sin.sin_zero, 0, sizeof(connect_addr.sin.sin_zero));
+        if (connect_addr.ss.ss_family == AF_INET)
+            memset(&connect_addr.sin.sin_zero, 0, sizeof(connect_addr.sin.sin_zero));
     }
 
     if (token[0] == '@') {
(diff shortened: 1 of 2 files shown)


lib/upipe-modules/upipe_udp.c
```diff
@@ -572,7 +572,7 @@ int upipe_udp_open_socket(struct upipe *upipe, const char *_uri, int ttl,
         else if (bind_addr.ss.ss_family != AF_UNSPEC) {
     normal_bind:
             if (bind(fd, &bind_addr.so, sockaddr_len) < 0) {
-                upipe_err(upipe, "couldn't bind");
+                upipe_err(upipe, "couldn't bind: %m");
```
cmassiot: Does it work ? I would think you need upipe_err_va for that to work. Also I thought %m was non-portable GNU extension.

funman: Hum good catch, yes it needs upipe_err_va.

%m is indeed a GNU extension (Glibc extension according to man printf) but we already use it, upipe_udp.c is even the biggest user.

I am pretty sure it doesn't work on Windows: https://msdn.microsoft.com/en-us/library/hf4y5e3w.aspx and it would print "%m" instead of the erorr string.

Don't know if it works on Darwin

cmassiot: So what do you want to do ? strerror_r seems to be a mess.

funman: Keep using %m for now

For Windows the error is in WSAGetLastError() and not errno anyway

cmassiot: So you must change it for upipe_err_va.

PR #224, opened 2016-10-11
Framers

funman:
A couple of framers fixes

### DIFF
--- a/lib/upipe-framers/upipe_s302_framer.c
+++ b/lib/upipe-framers/upipe_s302_framer.c
@@ -152,17 +152,22 @@ static void upipe_s302f_work(struct upipe *upipe, struct upump **upump_p)
 
     if (!ubase_check(uref_block_extract(upipe_s302f->next_uref,
                                         0, S302_HEADER_SIZE, header))) {
-        return;
+        uref_free(upipe_s302f->next_uref);
+        goto upipe_s302f_work_err;
     }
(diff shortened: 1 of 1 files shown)


lib/upipe-framers/upipe_h264_framer.c
```diff
@@ -1365,7 +1365,7 @@ static void upipe_h264f_build_flow_def(struct upipe *upipe)
 
     if (upipe_h264f->duration) {
         UBASE_FATAL(upipe, uref_clock_set_latency(flow_def,
-                         upipe_h264f->input_latency + upipe_h264f->duration))
+                         upipe_h264f->input_latency + upipe_h264f->duration * 2))
```
cmassiot: It gives me the impression that you are doubling the latency for all streams, not only interlaced.

funman: You're right.

I also see this code:

        case H264SEI_STRUCT_FRAME:
            UBASE_FATAL(upipe, uref_pic_set_progressive(uref))
            duration *= 2;
            break;

So no idea if this commit is correct. Also don't remember what it was trying to fix..

funman: It could have been a buggy stream which gave us a wrong latency.

I would drop that commit and keep only the s302 ones

funman: dropped

PR #238, opened 2016-10-28
Atomics

funman:
Use C11 atomics to make ThreadSanitizer happy


funman:
As discussed with nto, luajit needs to be told about uatomic_uint32_t

funman:
Closing, we need something that works when sharing code between C & C++

PR #246, opened 2016-11-20
Fix alignment for (un)pack10

kierank:


lib/upipe-hbrmt/upipe_pack10bit.c
```diff
@@ -222,6 +206,40 @@ static int upipe_pack10bit_check(struct upipe *upipe, struct uref *flow_format)
...
+{
+    struct uref *flow_format = uref_dup(request->uref);
+    UBASE_ALLOC_RETURN(flow_format);
+
+    uint64_t align;
+    if (!ubase_check(uref_block_flow_get_align(flow_format, &align)) || !align) {
+        uref_block_flow_set_align(flow_format, UBUF_ALIGN);
+        align = UBUF_ALIGN;
+    }
+
```
funman: there is extra whitespace

lib/upipe-hbrmt/upipe_pack10bit.c
```diff
@@ -246,6 +264,9 @@ static int upipe_pack10bit_set_flow_def(struct upipe *upipe, struct uref *flow_d
         return UBASE_ERR_ALLOC;
 
     uref_flow_set_def(flow_def_dup, "block.");
+    uref_block_flow_set_align(flow_def_dup, UBUF_ALIGN);
+    /* avx2 worst case, writes a full xmm register at offset + 10 */
+    uref_block_flow_set_append(flow_format, 10 + 16);
```
funman: should be flow_def_dup

lib/upipe-hbrmt/upipe_pack10bit.c
```diff
@@ -113,31 +115,13 @@ static bool upipe_pack10bit_handle(struct upipe *upipe, struct uref *uref,
     const char *def;
     if (unlikely(ubase_check(uref_flow_get_def(uref, &def)))) {
         upipe_pack10bit_store_flow_def(upipe, uref);
+        upipe_pack10bit_require_ubuf_mgr(upipe, flow_format);
```
kierank: We do that in v210enc...

lib/upipe-hbrmt/upipe_pack10bit.c
```diff
@@ -112,32 +114,14 @@ static bool upipe_pack10bit_handle(struct upipe *upipe, struct uref *uref,
...
 
-    size_t bytes;
-    uref_block_size(uref, &bytes);
-    if (!bytes || bytes % 32) {
-        upipe_err(upipe, "input buffer for pack10 must be padded to 32 bytes");
-        uref_free(uref);
-        return true;
-    }
-
-    if (unlikely(!upipe_pack10bit->ubuf_mgr)) {
```
funman: I think this check should stay with just a return false, because the ubuf_mgr has already been requested.

cmassiot: It's actually unneeded, because if there is no ubuf manager there will be no flow_def either, so we already return false.

kierank:
Do I need to do anything with this PR?

PR #247, opened 2016-11-22
Hbrmt dec

funman:
Depends on https://code.videolan.org/videolan/bitstream/merge_requests/6

This is a RFC since we're still waiting for hardware to test it properly.

Comments on how to integrate with rtp_decaps while still being fast are welcome :)
All we need from rtp_decaps is the marker bit.

### DIFF
--- a/lib/upipe-hbrmt/upipe_hbrmt_dec.c
+++ b/lib/upipe-hbrmt/upipe_hbrmt_dec.c
@@ -1,0 +1,473 @@
+/*
+ * Copyright (c) 2016 Open Broadcast Systems Ltd
+ *
+ * Authors: Kieran Kunhya
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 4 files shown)


cmassiot:
Perhaps we don't need to integrate rtp decaps as it is currently little more than a wrapper around biTStream. There is not so much code duplication.

lib/upipe-hbrmt/upipe_hbrmt_dec.c
```diff
@@ -0,0 +1,473 @@
...
+
+    const struct urational *fps = &frate_fps[frate - 0x10];
+    if (fps->num == 0) {
+        upipe_err_va(upipe, "Invalid hbrmt frate 0x%x", frate);
+        return UBASE_ERR_INVALID;
+    }
+
+    struct uref *flow_format = uref_dup(upipe_hbrmt_dec->flow_def);
+    uref_pic_flow_set_fps(flow_format, *fps);
+    upipe_hbrmt_dec->f = sdi_get_offsets(flow_format);
```
funman: Missing picture size (need to read FRAME from hbrmt header)

funman:
Will replace with a newer version

PR #249, opened 2016-11-29
DVEO ASI sink

funman:
This is an RFC.

Currently missing: 

- latency handling
- scope testing to make sure output rate is correct after dejittering input stream
- maybe options handling (buffers/bufsize)
- maybe another way to open the card (wait until first packet is received?)

Comments welcome!

### DIFF
--- a/lib/upipe-dveo/upipe_dveo_asi_sink.c
+++ b/lib/upipe-dveo/upipe_dveo_asi_sink.c
@@ -1,0 +1,766 @@
+/*
+ * Copyright (C) 2012-2015 OpenHeadend S.A.R.L.
+ * Copyright (C) 2016 Open Broadcast Systems Ltd
+ *
+ * Authors: Rafaël Carré
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject
+ * to the following conditions:
+ *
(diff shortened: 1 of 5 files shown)


lib/upipe-dveo/upipe_dveo_asi_sink.c
```diff
@@ -0,0 +1,765 @@
...
+
+    /* write header */
+    int size = 8;
+    uint8_t *header_write_ptr;
+    if (!ubase_check(ubuf_block_write(header, 0, &size, &header_write_ptr))) {
+        upipe_err(upipe, "could not write header");
+        ubuf_free(header);
+        return true;
+    }
+    memcpy(header_write_ptr, timestamp.timestamp, 8);
```
cmassiot: Are you sure the timestamp is in the native endianness ? If so, can you add a comment saying you checked ?

lib/upipe-dveo/upipe_dveo_asi_sink.c
```diff
@@ -0,0 +1,765 @@
...
+    }
+
+    unsigned long int mode;
+    snprintf(sys, sizeof(sys), sys_fmt, upipe_dveo_asi_sink->card_idx, "mode");
+    if (util_strtoul(sys, &mode) < 0) {
+        upipe_err(upipe, "Couldn't read buffer size");
+        goto error;
+    }
+
+    unsigned int packetsize;
```
cmassiot: Dead code.

funman: Which code is dead? github web UI and email do not show the same code and I'm not sure what's wrong here.

cmassiot: The packetsize calculation doesn't seem used anywhere.

cmassiot:
LGTM

funman:
Rebased all commits into one

PR #256, opened 2016-12-05
Bmdsrc

funman:
Allows to ask for v210 and s32


cmassiot:
perhaps /audio_bits= and /video_bits= would be clearer

PR #264, opened 2016-12-20
h264-framer: hint at progressive content

funman:


lib/upipe-framers/upipe_h264_framer.c
```diff
@@ -1371,6 +1371,10 @@ static void upipe_h264f_build_flow_def(struct upipe *upipe)
     upipe_h264f->annexb_requested =
         ubase_check(uref_h264_flow_get_annexb(upipe_h264f->flow_def_requested));
 
+    if (upipe_h264f->frame_mbs_only) {
```
cmassiot: Wrong place, it should be in activate_sps with the other flow def parameters.

PR #290, opened 2017-04-14
output helpers: check return value of unregister requests

funman:
assert since they should never fail


cmassiot:
duplicate of 7ffb19262b47f3b5d18bca861a1ec8afc70fcea6, closing

PR #293, opened 2017-05-30
upipe-av: include raw bits per sample in audio flow if known

funman:
This is set for example by the libavcodec s302m decoder


include/upipe/uref_sound_flow.h
```diff
@@ -50,6 +50,8 @@ UREF_ATTR_STRING_VA(sound_flow, channel, "s.channel[%" PRIu8"]",
 UREF_ATTR_SMALL_UNSIGNED(sound_flow, channels, "s.channels", number of channels)
 UREF_ATTR_SMALL_UNSIGNED(sound_flow, sample_size, "s.sample_size",
         size in octets of a sample of an audio plane)
+UREF_ATTR_SMALL_UNSIGNED(sound_flow, raw_sample_size, "s.raw_sample_size",
```
cmassiot: use a shorter string, it is expensive...

funman: Used s.sample_bits

PR #294, opened 2017-05-31
s337 framer: 20/24 bits support

funman:
Operate on sound data rather than block, it better matches the
sources of S337 data which are:
    - straight from a SDI sound buffer
    - MPEG-TS S302 data decoded by libavcodec

If block output is needed it can still be piped into upipe_tblk.

Do not do reallocation in the framer and do 1 frame in 1 frame out.

Notes:
    - unlike previous module, guard bands are currently kept as-is.
      they should probably be removed and added back at output
    - 16 bits AES in s32 is not handled

TODO:
    - timestamps are not currently handled.
    - modify bitstream a bit to accomodate for 20/24 bits
    - test?


funman:
Going with a new plan and a separate module

PR #306, opened 2017-07-21
Ffmpeg

funman:
A couple of FFmpeg API updates.

Refcounting seems to be mandatory since the introduction of the new decoding API (which we don't use yet).
Added refcounting for audio, but still missing subtitles refcounting.

AVCodecParameters stuff for avformat would like some testing.
Note: this breaks backward compatibility with FFmpeg < 3.1 so we might want to keep that commit around and apply it later


cmassiot:
Is it still true that it breaks compatibility with < 3.1 ?

funman:
No this is not true anymore, I removed the AVCodecParameters stuff (keeping it for later)

PR #328, opened 2017-09-25
Freetype

funman:
TODO: fuzzing

### DIFF
--- a/include/upipe-freetype/Makefile.am
+++ b/include/upipe-freetype/Makefile.am
@@ -1,0 +1,3 @@
+myincludedir = $(includedir)/upipe-freetype
+myinclude_HEADERS = \
+	upipe_freetype.h
--- a/lib/upipe-freetype/upipe_freetype.c
+++ b/lib/upipe-freetype/upipe_freetype.c
@@ -181,6 +181,7 @@ static void upipe_freetype_input(struct upipe *upipe, struct uref *uref, struct 
         uref_pic_flow_add_plane(flow_def, 1, 1, 1, "y8");
         uref_pic_flow_add_plane(flow_def, 2, 1, 1, "u8");
         uref_pic_flow_add_plane(flow_def, 2, 1, 1, "v8");
+        uref_pic_flow_add_plane(flow_def, 1, 1, 1, "a8");
         uref_pic_flow_set_hsize(flow_def, h);
         uref_pic_flow_set_hsize_visible(flow_def, h);
         uref_pic_flow_set_vsize(flow_def, v);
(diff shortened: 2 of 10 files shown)


funman:
Also need to look at harfbuzz (similar to pango)

lib/upipe-freetype/upipe_freetype.c
```diff
@@ -206,6 +206,7 @@ static void upipe_freetype_input(struct upipe *upipe, struct uref *uref, struct
     ubase_assert(ubuf_pic_plane_write(ubuf, "y8", 0, 0, -1, -1, &dst));
     uint8_t *dsta;
     ubase_assert(ubuf_pic_plane_write(ubuf, "a8", 0, 0, -1, -1, &dsta));
+    memset(dsta, 0, h * v);
```
cmassiot: It would be better if you fixed ubuf_pic_clear to also clear chroma.

lib/upipe-freetype/upipe_freetype.c
```diff
@@ -0,0 +1,374 @@
...
+        }
+        FT_Int y_max = y + bitmap->rows;
+        if (y_max > v) {
+            fprintf(stderr, "clipping y, %"PRIu64" < %d\n", v, y_max);
+            y_max = v;
+        }
+
+        for (FT_Int i = (x < 0 ? 0 : x); i < x_max; i++)
+            for (FT_Int j = (y < 0 ? 0 : y); j < y_max; j++)
+                dst[j*h + i] |= bitmap->buffer[(j - y) * bitmap->width + (i - x)];
```
cmassiot: You should not use h but the stride returned by ubuf_pic_plane_size.

lib/upipe-freetype/upipe_freetype.c
```diff
@@ -202,6 +202,10 @@ static void upipe_freetype_input(struct upipe *upipe, struct uref *uref, struct
 
     ubuf_pic_clear(ubuf, 0, 0, -1, -1, 0);
 
+    size_t stride_a, stride_y;
+    ubase_assert(ubuf_pic_plane_size(ubuf, "y8", &stride_y, NULL, NULL, NULL));
```
cmassiot: While ubase_assert() is OK for unit tests, asserting in a pipe is a bad pratice, and also if the user compiles without NDEBUG your command isn't executed. So you should replace those with UBASE_RETURN or UBASE_FATAL.

include/upipe-freetype/Makefile.am
```diff
@@ -0,0 +1,3 @@
+myincludedir = $(includedir)/upipe-freetype
+myinclude_HEADERS = \
+	upipe_freetype.h
```
cmassiot: I think you also need to link with libupipe.so as you use ubuf_pic_clear.

PR #329, opened 2017-09-25
upipe_sync: module to do A/V sync

funman:


lib/upipe-modules/upipe_sync.c
```diff
@@ -0,0 +1,1083 @@
...
+    uint64_t latency;
+    if (!ubase_check(uref_clock_get_latency(flow_def, &latency)))
+        latency = 0;
+
+    flow_def = uref_dup(flow_def);
+    if (!flow_def)
+        return UBASE_ERR_ALLOC;
+
+    // FIXME : estimated latency added by processing
+    latency += UCLOCK_FREQ / 25;
```
cmassiot: Shouldn't this be based on fps ?

funman: Yes

cmassiot: Do you intend to fix it ?

funman: Of course but will need some time to test it properly

cmassiot: Isn't it just a matter of adding * fps.den / fps.num ?

funman: Yes but I remember having problems with this so want to test for a while

cmassiot: Do you want to push it as is ?

funman: Yes I guess it doesn't harm

cmassiot: Can you squash it ?

PR #339, opened 2017-10-12
v210: do not depend on avutil headers

funman:


lib/upipe-v210/upipe_v210enc.c
```diff
@@ -133,16 +130,30 @@ UPIPE_HELPER_UBUF_MGR(upipe_v210enc, ubuf_mgr, flow_format, ubuf_mgr_request,
                       upipe_v210enc_unregister_output_request)
 UPIPE_HELPER_INPUT(upipe_v210enc, urefs, nb_urefs, max_urefs, blockers, upipe_v210enc_handle)
 
+static inline int clip(int a, int amin, int amax)
```
cmassiot: how about creating ubase_clip or so ?

funman: Good idea, the intreadwrite stuff could be useful too.

Should we put it all in ubase.h ?

cmassiot: I think so.

PR #349, opened 2017-10-13
blackmagic decklink sink

funman:


lib/upipe-blackmagic/sdi.c
```diff
@@ -0,0 +1,263 @@
...
+
+    dst[ANC_START_LEN + 9 + src_size + 3] = checksum ? 256 - checksum : 0;
+
+    dst[DC_POS] = cnt; // DC
+}
+
+static inline uint32_t to_le32(uint32_t a)
+{
+#ifdef UPIPE_WORDS_BIGENDIAN
+    return __builtin_bswap32(a);
```
cmassiot: I guess you will need a non-gcc implementation again.

lib/upipe-blackmagic/sdi.c
```diff
@@ -0,0 +1,263 @@
+#include <stdbool.h>
```
cmassiot: IIUC, there is nothing blackmagic-specific in this module, perhaps it should be moved to lib/upipe or upipe-modules ?

funman: True, although not sure what else could use it, maybe vanc filter?

cmassiot: Another SDI output module (dektec or deltacast for instance).

cmassiot:
Do you want to check it in as is ?

funman:
Why not, it works fine as-is.

We can always work on it later when someone add output for another card

cmassiot:
So can you squash it ?

PR #361, opened 2017-10-27
Apple clang now knows about builtin_sse("ssse3")

funman:


cmassiot:
can you fix the conflict ?

PR #364, opened 2017-11-03
Dvbsub

funman:


funman:
I think multiple subs with the same pts will not work with subpic_schedule

funman:
Now it can, but then due to blit design the last uref input supercedes all the previous ones :(

funman:
Did it differently, no need to change the other modules anymore

PR #365, opened 2017-11-06
add zoneplate source

JDarnley:


lib/upipe-filters/Makefile.am
```diff
@@ -10,7 +10,10 @@ libupipe_filters_la_SOURCES = \
 	upipe_audio_bar.c \
 	upipe_audio_graph.c \
 	ebur128/ebur128.c \
-	ebur128/ebur128.h
+	ebur128/ebur128.h \
+	upipe_zoneplate_source.c \
+	videotestsrc.c \
```
cmassiot: If it is a file coming from another project, I think it should be in a subdirectory, same as http_parser.

JDarnley: Should the subdir be based on the originating project (gstreamer or specifically gst-plugins-base) or the file I've copied (videotestsrc)?  Would you prefer I didn't squash that change into the commit?

quarium: Maybe zoneplate for the directory name?

PR #375, opened 2017-11-16
Change x86 assembler define for the C preprocessor

JDarnley:
I thought this was a good idea to reduce the changes I need to do to add checkasm.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -91,7 +91,7 @@ then
     fi
 fi
 
-AM_CONDITIONAL(HAVE_X86_ASM, test -n "${NASM}" -a -n "${NASMFLAGS}")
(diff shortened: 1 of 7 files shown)


JDarnley:
I need to know if the assembler is available for checkasm.  It has a 4 functions of its own in an asm file.  Using these obviously depends on them being built.  As does the list of functions to test.  They need to be present also.

I guess I could simply add a cflag to the makefile like you have done in hbrmt and v210.  Is that the autotools way?  Is that the preferred way?

funman:
There is no preferred way, but IIUC you can not include config.h in an asm file, however -D would work.

JDarnley:
Sorry if I wasn't clear.  Only in C and the Makefile do I need to know.  If the assembler is run it is clearly available.

funman:
OK then this works

PR #376, opened 2017-11-16
Docs

Jude188:
Made clear that the upipe_foo_alloc() function requires a flowdef to define input as an argument.


kierank:
Trailing whitespace

PR #381, opened 2017-11-30
Dvbcsa

funman:


cmassiot:
Can you squash it ?

funman:
I can but why ? These are separate logical changes IMO

cmassiot:
Ok, didn't actually read them :).

PR #382, opened 2017-12-01
Copy FFmpeg's checkasm test utility

JDarnley:
For those unaware this utility is a great platform on which to test SIMD functions.  It iterates over a list of functions and calls each after forcing CPU feature flags to be each one available on the host.  Each function in the list called is a specially written function that: sets up data structures needed to run a reference and tested functions; creates input data; runs both the reference and tested functions; checks whether the output is correct, usually whether it is identical to the reference output; finally it reports the status to the main program.  It also provides an option to benchmark the functions

This PR is intended to get comments on adopting this into Upipe.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -332,6 +332,7 @@ AC_CONFIG_FILES([Makefile
                  x86/Makefile
                  x86/config.asm
                  tests/Makefile
+                 tests/checkasm/Makefile
                  examples/Makefile
                  luajit/Makefile])
 AC_OUTPUT
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1,6 +1,10 @@
 LOG_COMPILER = $(srcdir)/valgrind_wrapper.sh
 AM_LOG_FLAGS = $(srcdir)
 
+if HAVE_AVUTIL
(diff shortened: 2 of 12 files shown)


JDarnley:
One thing I wanted to ask about was whether it would be a good idea to to change the layout of the tests and files.  Now they are aligned with the source files of the functions tested.  For example v210enc tests the v210 packing functions from v210enc.asm.  It creates both planar8 and planar 10 input data for it.

I want to suggest changing the layout so that one testing function only has to create one type of reference data.  As an example v210enc has to create v210 data and only has to compare the output buffer with the reference.  I would suggest renaming this to v210_input.  Likewise, functions that read sdi input data should go in sdi_input, uyvy reading functions go in uyvy_input.

I think this made a lot of sense with many more functions to test (5 input * 4 output) which were spread over a couple more files.

nto:
Can you squash all bugfix commits into previous ones? Basically I think there should be one commit for the checkasm infrastructure, and commits for adding v210 and sdi enc/dec support, and that's all.

About the layout of the test files, if you think it makes more sense to change it like you suggested, go for it.

PR #392, opened 2017-12-18
Move some C functions into dedicated files

JDarnley:
At the request of @nto I have cherry-picked these commits from the checkasm pull request.

### DIFF
--- a/lib/upipe-hbrmt/Makefile.am
+++ b/lib/upipe-hbrmt/Makefile.am
@@ -2,6 +2,8 @@ lib_LTLIBRARIES = libupipe_hbrmt.la
 
 libupipe_hbrmt_la_SOURCES = upipe_pack10bit.c \
     upipe_unpack10bit.c \
+    sdidec.c \
+    sdidec.h \
     $(NULL)
 
 libupipe_hbrmt_la_CPPFLAGS = -I$(top_builddir)/include -I$(top_srcdir)/include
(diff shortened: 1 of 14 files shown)


cmassiot:
Your license header only speaks of ffmpeg, shouldn't it be changed to Upipe ? Especially since some functions cannot possibly originate from ffmpeg.

JDarnley:
Hm.  You might be right.  I just duplicated the existing files and removed the parts as needed.  The licenses on the new files in upipe-hbrmt could be changed, maybe.  Do you want me to do that?

cmassiot:
Yes please

JDarnley:
@cmassiot @nto  I made the license header changes a few days ago.  Do you have any more comments?

nto:
v210enc.c and v210dec.c still shows the ffmpeg license notice. Apparently it's the case for some older files as well, so maybe they all need to be changed in a separate pull request. It's definitely misleading to say "You should have received a copy of the GNU Lesser General Public License along with FFmpeg" in these files.

JDarnley:
Yes, I didn't change those files because of this line in the header "This file is based on the implementation in FFmpeg".  FFmpeg does have a v210 codec so these functions are likely to be derived from them.

I only moved the functions from one file to another so I shouldn't be changing what license they have (in my understanding of these things).  I probably shouldn't change the license headers on the 10-bit sdi packing either but you could argue that the header was wrong since the beginning in commit 9a54a727

If you think I should change the named library (FFmpeg) in the new v210 files I will.

nto:
The license and authorship informations is a real mess right now in Upipe. We can apply this pull request now so you can move on to the checkasm patch. We will review the license notices situation later.

PR #397, opened 2018-01-05
Add extern C to public headers

studoret:
Note that I placed '#ifdef __cplusplus' after '#include' directives because I consider that each header file takes care of itself.
By the way, Happy new year !

### DIFF
--- a/include/upipe/udict.h
+++ b/include/upipe/udict.h
@@ -596,7 +596,7 @@ static inline int udict_set_opaque_from_hex(struct udict *udict,
 {
     size_t size = (strlen(value) + 1) / 2;
     uint8_t v[size];
-    int i;
+    size_t i;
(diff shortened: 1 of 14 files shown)


include/upipe/udict.h
```diff
@@ -596,7 +596,7 @@ static inline int udict_set_opaque_from_hex(struct udict *udict,
 {
     size_t size = (strlen(value) + 1) / 2;
     uint8_t v[size];
-    int i;
+    uint32_t i;
```
quarium: Maybe size_t is more appropriate

studoret: I agree.

nto:
Could you squash the two sign-compare commits into a single one?

studoret:
Done

nto:
Thanks!

PR #404, opened 2018-01-11
upipe-av: set picture size before opening decoder

nto:


cmassiot:
No consensus on this one.

PR #408, opened 2018-01-23
Nasm

funman:


funman:
@JDarnley @kierank

JDarnley:
Do we want to check for a new enough version of nasm/yasm to ensure that they support AVX2?  When they're too old they should return an error when building and they reach an instruction they don't know.  So it probably isn't worth it just to get a nice message out of configure.

PR #409, opened 2018-01-23
Fix avutil dependency

funman:
These files depend on avutil headers, and build will fail if it libavutil is not installed in default prefix


examples/Makefile.am
```diff
@@ -35,6 +35,7 @@ endif
 if HAVE_AUDIOTOOLBOX
 uplay_LDADD += $(UPIPEOSX_LIBS)
 endif
+uplay_CFLAGS = $(AVUTIL_CFLAGS)
```
cmassiot: Should be $(UPIPEAV_CFLAGS) for consistency.

funman: Do you want me to do UPIPEAV_CFLAGS = $(AVUTIL_CFLAGS) ?

cmassiot: UPIPEAV_CFLAGS should already exist.

nto: It seems there is no way to use `libupipe_av_la_CFLAGS` in examples/Makefile.am, so your suggestion (`UPIPEAV_CFLAGS = $(AVUTIL_CFLAGS)`) looks fine to me.

PR #428, opened 2018-02-06
Arq

funman:


lib/upipe-filters/upipe_rtp_feedback.c
```diff
@@ -562,6 +562,7 @@ static void upipe_rtpfb_send_rr(struct upipe *upipe)
 
     /* Header */
     rtcp_set_rtp_version(buf);
+    buf[0] |= 1; /* RC */
```
cmassiot: Shouldn't it be a rtcp_set_XXX ?

funman: Yes

PR #434, opened 2018-02-21
ts_demux: fix restart of ES on PMT change

cmassiot:
Bug introduced in 424c529c7c1490c12056dccb2d2620a9810702ff


funman:
I would prefer if you split the changes in multiple commits next time

PR #442, opened 2018-03-01
Upipe zoneplate

quarium:


JDarnley:
Let me know if I understand this correctly.  You are moving drawing zoneplate patterns into its own pipe.  I guess so you can draw onto any picture you already have.  Then I think you are modifying the zoneplate source pipe to create pictures another way and pass those to the drawing pipe.  I guess that "other way" is the void source pipe.  Is that all correct?

It sounds like a worthwhile change.  I think you should be more verbose in the commit message for it to help idiots like me understand it.  Do you already have a use for it?  Is any change to code already using zoneplate source needed?  I will try testing it in one of utilities.  Will void source waste time clearing the picture it creates?

Do you want me to read the code more thoroughly?

quarium:
Yes this is exactly what I've done. But zoneplate pipe creates its own picture and attach it to the input uref from void source (not a picture, only a udict), so no need to clear. And yes we have a use case =).
I'll add some explanation in the commit message if you want.
I'd like you to check that zoneplate source works exactly as before, it's supposed to. And if you have some time, may be read the code to give me your opinion on it.

JDarnley:
Okay.  I have just merged your changes into our code (the only conflicts were in examples/makefile.am) and built things.  I needed a small change to the program because you renamed the allocation function but that was simple thanks to the different header defines.

Running the program I get this
```
$ ./vc2-send temp.vc2
  debug: freezing upump manager (1)
  debug: [zoneplate] throw ready event
  error: [zoneplate] fail to allocate source pipe
  debug: [zoneplate] throw dead event
  debug: thawing upump manager (0)
vc2-send: vc2-send.c:191: main: Assertion `pipe' failed.
Aborted (core dumped)
```

I am not sure where to start diagnosing the issue other than saying the message comes from your new code.

Issue #414, opened 2018-01-29
examples/{u,glx}play: incorrect stride on some sources


tmatth:
To reproduce, run `./uplay ed.480p.ts` to get this type of output:
![uplay-ed](https://user-images.githubusercontent.com/66862/35523519-09332f90-04ed-11e8-91ee-2706233378c8.png)

or similarly (but in monochrome) for `./glxplay ed.480p.ts`:
![glxplay-ed](https://user-images.githubusercontent.com/66862/35523563-229e24ee-04ed-11e8-8263-4f1f5f0d79d9.png)

Sample:
http://people.videolan.org/~tmatth/ed.480p.ts

funman:
It works with uplay -s 856x480 although the file is 854x480

funman:
glxplay is different because it uses rgb24 rather than rgb15

funman:
```
diff --git a/lib/upipe-av/upipe_avcodec_decode.c b/lib/upipe-av/upipe_avcodec_decode.c
index bf674ca8..46b4c969 100644
--- a/lib/upipe-av/upipe_avcodec_decode.c
+++ b/lib/upipe-av/upipe_avcodec_decode.c
@@ -295,8 +295,8 @@ static int upipe_avcdec_get_buffer_pic(struct AVCodecContext *context,
     }
 
     UBASE_FATAL(upipe, uref_pic_flow_set_align(flow_def_attr, align))
-    UBASE_FATAL(upipe, uref_pic_flow_set_hsize(flow_def_attr, context->width))
-    UBASE_FATAL(upipe, uref_pic_flow_set_vsize(flow_def_attr, context->height))
+    UBASE_FATAL(upipe, uref_pic_flow_set_hsize(flow_def_attr, context->coded_width))
+    UBASE_FATAL(upipe, uref_pic_flow_set_vsize(flow_def_attr, context->coded_height))
     UBASE_FATAL(upipe, uref_pic_flow_set_hsize_visible(flow_def_attr, context->width))
     UBASE_FATAL(upipe, uref_pic_flow_set_vsize_visible(flow_def_attr, context->height))
     struct urational fps;
```

funman:
@cmassiot ?

tmatth:
That change works as expected here.

nto:
Additionally, shouldn't upipe-gl use hsize_visible and vsize_visible somewhere to crop the visible surface?

cmassiot:
[...]

PR #450, opened 2018-03-14
upipe-av: use coded_width / height

funman:
Fixes issue #414


funman:
Closing due to comments in issue, there is a new PR to respect stride in gl

PR #451, opened 2018-03-14
Use github repo for bitstream

cmassiot:


cmassiot:
code.videolan.org is actually the official repo

PR #456, opened 2018-03-19
gl_sink: respect stride

funman:
Fixes issue #414


tmatth:
This doesn't fix #414 for me.

tmatth:
I can confirm that the revised patch https://github.com/Upipe/upipe/pull/456/commits/6f61e56a5f247a308224e75217eb3e571684cbf6 fixes both uplay and glxplay

PR #459, opened 2018-03-19
alsa: Reduce buffer size to something allowed on Vx222e

kierank:


lib/upipe-alsa/upipe_alsa_sink.c
```diff
@@ -380,7 +380,7 @@ static bool upipe_alsink_open(struct upipe *upipe)
     upipe_alsink->period_duration = (uint64_t)frames_in_period *
                                     UCLOCK_FREQ / upipe_alsink->rate;
 
-    snd_pcm_uframes_t buffer_size = frames_in_period * 4;
+    snd_pcm_uframes_t buffer_size = frames_in_period * 2;
```
cmassiot: I don't think this is consistent with what's coming later, for instance line 405.

PR #460, opened 2018-03-20
alsa: add an option to set the capture period

funman:
it needs to be set before opening the device

Note #459 is no longer needed with this

### DIFF
--- a/lib/upipe-alsa/upipe_alsa_source.c
+++ b/lib/upipe-alsa/upipe_alsa_source.c
@@ -126,6 +126,9 @@ struct upipe_alsource {
     /** poll fd **/
     struct pollfd pfd;
 
+    /** period */
+    uint64_t period;
+
     /** pipe acting as output */
     struct upipe *output;
     /** flow definition packet */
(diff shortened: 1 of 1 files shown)


lib/upipe-alsa/upipe_alsa_source.c
```diff
@@ -184,8 +187,9 @@ static struct upipe *upipe_alsource_alloc(struct upipe_mgr *mgr,
     upipe_alsource->rate = 48000;
     upipe_alsource->channels = 2;
     upipe_alsource->format = SND_PCM_FORMAT_FLOAT;
-    upipe_alsource->uri = strdup(DEFAULT_DEVICE);
+    upipe_alsource->uri = NULL;
```
cmassiot: Is this an intended change ?

funman: Yes but it seems incomplete, will fix it

funman: Hm actually it seems good as-is.

upipe_alsource_open() is only called from set_uri, and set_uri needs uri to be set, effectively overwriting this first assignement.

PR #465, opened 2018-04-17
Clang static analyzer

funman:
Fix some issues noticed by clang static analyzer

Full report on http://chui-pas.net/~fun/2018-04-17-144958-11229-1/


lib/upipe-framers/upipe_s337_framer.c
```diff
@@ -410,7 +410,7 @@ static int upipe_s337f_set_flow_def(struct upipe *upipe, struct uref *flow_def)
         return UBASE_ERR_ALLOC;
     }
 
-    flow_def = upipe_s337f_store_flow_def_input(upipe, flow_def_dup);
+    upipe_s337f_store_flow_def_input(upipe, flow_def_dup);
```
cmassiot: You need to do something with the built flow def otherwise it's a memleak. Usually you store_flow_def() it.

PR #469, opened 2018-04-20
upipe_autoin: add automatic inner pipe allocator

quarium:
This pipe automatically allocates inner pipes when input flow def
changes based on the pipe managers added to the auto inner pipe manager.
Idem pipe could be use at the end to allow passtrough.


lib/upipe-modules/upipe_auto_inner.c
```diff
@@ -0,0 +1,388 @@
...
+static int _upipe_autoin_mgr_add_mgr(struct upipe_mgr *mgr,
+                                     const char *name,
+                                     struct upipe_mgr *inner_mgr)
+{
+    struct upipe_autoin_mgr *upipe_autoin_mgr = upipe_autoin_mgr_from_mgr(mgr);
+
+    if (unlikely(inner_mgr == NULL))
+        return UBASE_ERR_INVALID;
+
+    if (unlikely(upipe_autoin_mgr_find(mgr, inner_mgr)))
```
cmassiot: You should only allow this if urefcount_single(&mgr->refcount). Otherwise there is a race condition with already allocated pipes. See upipe_filter_format.

lib/upipe-modules/upipe_auto_inner.c
```diff
@@ -0,0 +1,388 @@
...
+ * @param mgr auto inner pipe manager
+ * @param inner_mgr inner manager to delete
+ */
+static void _upipe_autoin_mgr_del_mgr(struct upipe_mgr *mgr,
+                                      struct upipe_mgr *inner_mgr)
+{
+    struct upipe_autoin_mgr_item *item =
+        upipe_autoin_mgr_find(mgr, inner_mgr);
+    if (item) {
+        ulist_delete(&item->uchain);
```
cmassiot: Same comment here (ulist_delete is not thread-safe).

PR #470, opened 2018-04-20
upipe_v210: fix compilation without asm

quarium:


kierank:
lgtm

quarium:
The diff in hbrmt is useless, i'll remove it

funman:
And the changes in configure.ac, isn't it separate too ?

quarium:
Without the change in configure.ac HAVE_X86ASM is always defined

funman:
Ah ok, that's fine then I had not fully understood the issue

PR #485, opened 2018-05-15
Split lines and separate fields

funman:


lib/upipe-modules/upipe_row_split.c
```diff
@@ -0,0 +1,357 @@
...
+        upipe_throw_error(upipe, UBASE_ERR_INVALID);
+        uref_free(uref);
+        return true;
+    }
+
+    uint64_t vsize_slice = upipe_row_split->slice_height;
+    if (vsize_slice > vsize)
+        vsize_slice = vsize;
+
+    for (uint64_t done = 0; done < vsize; done += vsize_slice) {
```
cmassiot: I don't understand why you don't do simply:
new_ubuf = ubuf_dup(ubuf);
ubuf_pic_resize(new_ubuf, 0, done, -1, vsize_slice);
done += vsize_slice;

lib/upipe-modules/upipe_row_split.c
```diff
@@ -0,0 +1,357 @@
...
+        struct ubuf *ubuf = ubuf_pic_alloc(upipe_row_split->ubuf_mgr, hsize, vsize_slice);
+        if (!ubuf)
+            abort();
```
cmassiot: Do not call abort() from pipes.

lib/upipe-modules/upipe_row_split.c
```diff
@@ -0,0 +1,357 @@
...
+            // TODO: dst stride != src stride
+
+            ubase_assert(uref_pic_plane_unmap(uref, chroma, 0, done, -1, vsize_slice));
+            ubase_assert(ubuf_pic_plane_unmap(ubuf, chroma, 0, 0, -1, -1));
+
+        }
+
+        struct uref *uref_slice = uref_fork(uref, ubuf);
+
+        uref_attr_set_unsigned(uref_slice,
```
cmassiot: Please use the macro UREF_ATTR_UNSIGNED in header.

lib/upipe-modules/upipe_separate_fields.c
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    bool has_progressive_attr = ubase_check(uref_pic_get_progressive(uref));
+    bool has_tff_attr         = ubase_check(uref_pic_get_tff(uref));
+
+    if (has_progressive_attr)
+        upipe_warn(upipe, "picture marked as progrssive, separating fields anyway");
+
+    /* first field */
+    struct ubuf *ubuf = ubuf_pic_alloc(ctx->ubuf_mgr, width, height / 2);
```
cmassiot: Technically it could work with zero copy by changing the origin or end, and changing the stride. Changing the stride would require patching the manager but it should be easy.

lib/upipe-modules/upipe_row_split.c
funman: Ideally these would be used by more modules in the future, not sure in which common location I should put them

cmassiot: uref_pic ?

lib/upipe-modules/upipe_separate_fields.c
funman: OK need to do this at some point, I wonder if I should split the PR in 2 modules then

JDarnley:
Before any of this gets committed can I make a quick comment about what term Upipe will/should use to refer to a part of a frame.  I guess it does depends on Upipe accepting this feature.

The row_split pipe uses the term "slice" internally and in a couple of log messages.  That file's comment description has "split pictures into a series of rows".  We have used slice in some of our other code too when using a part of a frame.  This has led to confusion on occasion when also discussing codec/format features also called slices.  In some new code I have for FFmpeg's draw_horiz_band feature I have used "chunk".

Some alternatives that could be used are also used by other codecs/formats in specific ways.
* slice: as mentioned above it is used in many codecs as a technical feature.  I don't mind sticking to this term if others agree it should be used.
* fragment: Dirac/VC-2 has this as a bitstream packing feature.  I think it is also used by the pseudo-streaming MP4 files used on websites.
* chunk: I don't know this term being used by anything.
* subpicture: I can, maybe, see this being confused with subtitles and I would not be surprised to learn that some codec uses it in a technical manner.
* band, horizontal_band: Dirac/VC-2 use (sub)band when speaking about its wavelet transforms.  Clearly ties with FFmpeg's feature.

I just wanted to bring this to everyone's attention, offer and ask for some suggestions, and see if we can agree on a term.

This isn't as quick as I thought it would be.

lib/upipe/ubuf_pic_common.c
```diff
@@ -288,6 +288,53 @@ int ubuf_pic_common_check_skip(struct ubuf_mgr *mgr, int hskip, int vskip)
...
+int ubuf_pic_common_split_fields(struct ubuf *ubuf, struct ubuf **odd,
+        struct ubuf **even)
+{
+    *odd = ubuf_dup(ubuf);
+    if (!*odd)
+        return UBASE_ERR_ALLOC;
+
+    *even = ubuf_dup(ubuf);
+    if (!*odd) {
+        ubuf_free(*odd);
```
JDarnley: Do you need to do anything about the value stored in odd?  Do you need to set it to NULL or anything?

cmassiot: I guess he meant if (!even)

funman: No if the function returns an error the contents are undefined

funman: @cmassiot nice catch, fixed it

include/upipe/uref_pic.h
```diff
@@ -171,6 +172,27 @@ static inline int uref_pic_plane_clear(struct uref *uref,
...
+{
+    int ret = UBASE_ERR_ALLOC;
+
+    *odd = uref_dup_inner(uref);
+    *even = uref_dup_inner(uref);
+
+    if (*odd && *even)
+        ret = ubuf_split_fields(uref->ubuf, &(*odd)->ubuf, &(*even)->ubuf);
+
+end:
```
JDarnley: Is this label a leftover of something or did you leave something out when committing?  I think you've missed freeing odd or even should the allocation of the other fail.

PR #486, opened 2018-05-16
upipe-x265: add auto bit depth selection

nto:


lib/upipe-x265/upipe_x265.c
```diff
@@ -230,6 +243,132 @@ UPIPE_HELPER_UBUF_MGR(upipe_x265, ubuf_mgr, flow_format, ubuf_mgr_request,
...
+        params->fpsDenom = fps.den;
+    }
+
+    params->vui.aspectRatioIdc = upipe_x265->aspect_ratio_idc;
+    if (params->vui.aspectRatioIdc == X265_EXTENDED_SAR) {
+        params->vui.sarWidth = upipe_x265->sar_width;
+        params->vui.sarHeight = upipe_x265->sar_height;
+    }
+    params->vui.bEnableOverscanInfoPresentFlag = upipe_x265->overscan;
+    params->vui.bEnableOverscanAppropriateFlag = upipe_x265->overscan;
```
cmassiot: Actually, the correct behaviour should be:
 - if the attribute is not present, InfoPresentFlag = false
 - if the attribute is present, InfoPresentFlag = true; AppropriateFlag = overscan

PR #492, opened 2018-06-06
upipe_blit: add sub pipe flush control command

quarium:


include/upipe-modules/upipe_blit.h
```diff
@@ -88,7 +88,9 @@ enum upipe_blit_sub_command {
     /** gets the z-index (int *) */
     UPIPE_BLIT_SUB_GET_Z_INDEX,
     /** sets the z-index (int) */
-    UPIPE_BLIT_SUB_SET_Z_INDEX
+    UPIPE_BLIT_SUB_SET_Z_INDEX,
+    /** flush (void) */
+    UPIPE_BLIT_SUB_FLUSH
```
cmassiot: Why not use the standard control command UPIPE_FLUSH ?

quarium: Fixed

PR #499, opened 2018-06-12
rtp_fec: Add various improvements

kierank:
Mainly helps if reorder pipe is placed in front of FEC

### DIFF
--- a/lib/upipe-ts/upipe_rtp_fec.c
+++ b/lib/upipe-ts/upipe_rtp_fec.c
@@ -642,16 +642,21 @@ static void upipe_rtp_fec_main_input(struct upipe *upipe, struct uref *uref)
         return;
     }
 
+    /* Difference between last received sequence number and current sequence number */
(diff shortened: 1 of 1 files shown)


lib/upipe-ts/upipe_rtp_fec.c
```diff
@@ -551,8 +551,7 @@ static void clear_fec(struct upipe *upipe)
     upipe_rtp_fec->last_seqnum = UINT32_MAX;
     upipe_rtp_fec->latency = 0;
 
-    memset(upipe_rtp_fec->recent, 0xff, 2 * upipe_rtp_fec->rows *
-            upipe_rtp_fec->cols * sizeof(*upipe_rtp_fec->recent));
+    memset(upipe_rtp_fec->recent, 0xff, sizeof(upipe_rtp_fec->recent));
 }
```
funman: This is not needed, all accesses to ->recent are done modulo 2 * rows * col already

PR #501, opened 2018-06-14
fec: check payload type

funman:
If a packet has a payload type different from the first packet,
forward it to the next pipe immediately.
This lets us include not only RTCP but also retransmits which are
multiplexed together with the original RTP stream


funman:
Another possibility is to fix the payload type on pipe alloc rather than wait first packet

cmassiot:
Then shouldn't the user provide the correct PT that should be processed ?

cmassiot:
LGTM.

PR #506, opened 2018-07-23
new pipe: row join

JDarnley:
This adds a new pipe to join chunks into a new picture.  It does the reverse operation of row_split that @funman wants to add in PR #485.

I do need to add a test for it.


cmassiot:
LGTM.

JDarnley:
Oh, I guess I should say that I rebased, squashed all the commits together, and pushed.

PR #508, opened 2018-08-03
Add support for FFmpeg's draw_horiz_band feature

JDarnley:
Very experimental but does work together with row_join (#506) and some unpublished changes to uplay.

### DIFF
--- a/lib/upipe-av/upipe_avcodec_decode.c
+++ b/lib/upipe-av/upipe_avcodec_decode.c
@@ -1087,18 +1087,12 @@ static void draw_horiz_band(AVCodecContext *avctx, const AVFrame *frame,
         return;
     }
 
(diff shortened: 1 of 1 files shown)


lib/upipe-av/upipe_avcodec_decode.c
```diff
@@ -297,6 +297,9 @@ static int upipe_avcdec_get_buffer_pic(struct AVCodecContext *context,
         return -1;
     }
 
+    if (frame->interlaced_frame)
+        UBASE_FATAL(upipe, uref_attr_set_void(flow_def_attr, NULL, UDICT_TYPE_VOID, "sepfields"));
```
cmassiot: Please use a macro to define a new attribute.

JDarnley:
I think we eventually want finish work on this but we have so many more features in progress.  I am going to move it to a draft.

PR #514, opened 2018-08-09
Two things are certain in life, death and O(packet) logging.

kierank:


cmassiot:
Note that if you want to improve the performance of your pipe, you should #ifdef the line. Because even if uprobe_stdio is configured not to print upipe_dbg, it will go up the long line of probes.

PR #530, opened 2018-10-23
S12M timecode support

januaryjp:
Source from lavc, Source/Sink for Blackmagic

### DIFF
--- a/lib/upipe-av/upipe_avcodec_decode.c
+++ b/lib/upipe-av/upipe_avcodec_decode.c
@@ -1093,6 +1093,11 @@ static void upipe_avcdec_output_pic(struct upipe *upipe, struct upump **upump_p)
     if (side_data)
         uref_pic_set_cea_708(uref, side_data->data, side_data->size);
 
+    side_data = av_frame_get_side_data(frame, AV_FRAME_DATA_S12M_TIMECODE);
+    if (side_data) {
+        uref_pic_set_s12m(uref, side_data->data, side_data->size);
+    }
+
     /* various time-related attributes */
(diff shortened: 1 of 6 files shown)


cmassiot:
I guess we do not have a recent enough version of ffmpeg to build it ?

januaryjp:
Yes, I guess I would need to add some sort of guard for ffmpeg version?

cmassiot:
Yes it would be lovely.

nto:
Just add a check for the needed ffmpeg symbols in configure, so you don't need to depend on a specific release.

lib/upipe-av/upipe_avcodec_decode.c
```diff
@@ -1093,6 +1093,11 @@ static void upipe_avcdec_output_pic(struct upipe *upipe, struct upump **upump_p)
     if (side_data)
         uref_pic_set_cea_708(uref, side_data->data, side_data->size);
 
+    side_data = av_frame_get_side_data(frame, AV_FRAME_DATA_S12M_TIMECODE);
+    if (side_data) {
+        uref_pic_set_s12m(uref, side_data->data, side_data->size);
+    }
+
```
nto: What about using the `AV_FRAME_DATA_S12M_TIMECODE` macro to check for availability : 
```
#ifdef AV_FRAME_DATA_S12M_TIMECODE
     side_data = av_frame_get_side_data(frame, AV_FRAME_DATA_S12M_TIMECODE);
     if (side_data) {
         uref_pic_set_s12m(uref, side_data->data, side_data->size);
     }
#endif
```

cmassiot:
Superseded by #991.

PR #532, opened 2018-10-26
Hls playlist

quarium:


cmassiot:
Please check the commit messages, they do not seem accurate.

PR #535, opened 2018-10-31
upipe-ts: CAT decoder for BISS-CA

funman:
This is the first stone for a BISS-CA decoder

TODO:
- decode CA descriptor in PMT (pointing to ECM)
- decode EMM
- decode ECM
- decrypt session key in EMM
- decrypt session words in ECM
- decrypt ES
- maybe implement mode E, although in the real world mode E uses BISS 1 / DVB-CSA1 rather than BISS 2 / AES-128 (aka DVB-CISSA)


cmassiot:
LGTM

PR #543, opened 2018-11-21
upipe_sync: Add probe for missing picture

kierank:


kierank:
ah yes forgot to update docs.

PR #545, opened 2018-12-05
upipe-av: remove generated file on distclean

nto:


funman:
I'm Rafaël Carré, and I approve this message^Wpull request.

PR #552, opened 2018-12-18
dvbcsa_decrypt: odd/even session words

funman:
This adds support for decrypting streams with odd/even keys.

Encryption is missing, we need a mechanism to toggle between odd/even keys (manual control ?)

### DIFF
--- a/examples/ts_encrypt.c
+++ b/examples/ts_encrypt.c
@@ -141,7 +141,7 @@ static void usage(const char *name)
 {
     fprintf(stderr, "%s [options] <input> <output>\n"
             "\t-v   : be more verbose\n"
-            "\t-k   : set BISS key\n"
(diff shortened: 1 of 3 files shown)


examples/ts_encrypt.c
```diff
@@ -188,7 +188,15 @@ int main(int argc, char *argv[])
                     log_level = UPROBE_LOG_DEBUG;
                 break;
             case OPT_KEY:
-                key = optarg;
+                if (!key[0]) {
+                    key[0] = optarg;
+                } else if (!key[1]) {
+                    key[1] = optarg;
+                } else {
+                    fprintf(stderr, "key specified more than twice");
```
nto: Missing `\n`?

quarium:
You can update the authors list

PR #555, opened 2019-01-11
H264

funman:


funman:
We are still keeping a uref, it's just the first of N urefs until a buffer is getting output.
I don't see how a regression is possible as after we set stored uref to NULL the upipe_avcdec_input call is done, and there's gonna be a uref with the next one.

kierank:
ping

cmassiot:
Is this pull request still necessary?

PR #556, opened 2019-01-15
dvbcsa_decrypt: fix odd/even keys

funman:


lib/upipe-dvbcsa/upipe_dvbcsa_decrypt.c
```diff
@@ -404,11 +404,19 @@ static void upipe_dvbcsa_dec_input(struct upipe *upipe,
     uint16_t pid = ts_get_pid(ts_header);
     uref_block_peek_unmap(uref, 0, buf, ts_header);
 
-    bool odd = scrambling & (1 << 1);
+    bool odd, valid = false;
+    switch (scrambling) {
+        case 2:
```
cmassiot: There should be a define in bitstream for this.

funman: OK if I put it in mpeg/ts.h even though it's not technically part of iso 13818 ?

cmassiot: I thought it was, but indeed it is so common we can put it there...

PR #584, opened 2019-04-05
rist: add more commands for stats

nto:
upipe_rtcpfb_get_stats
upipe_rtcpfb_reset_stats
upipe_rtpfb_reset_stats
upipe_rtpfb_get_rtt


funman:
I would rather do like rtpd and reset all counters to 0 when they are read, for consistency.

What do you think ?

Also cc @kierank

nto:
`rtpfb` only resets a few counters, not all of them (eg: `lost`), that's why I had to add the `reset_stats` commands. Do you mean it should be changed to reset all counters instead ?

funman:
I think so yes, sorry to have missed that

nto:
Done. Much simpler that way.

PR #589, opened 2019-04-12
Fix rtcp and rtp feedback

quarium:


funman:
Which problems does it fix though ?

quarium:
- Reallocate the upump when the upump manager changes
- Handle register/unregister requests
- Forward the upump downstream so it can be blocked by sink
- Handle attach uclock
- Remove unused code + cosmetic

PR #596, opened 2019-04-18
upipe-modules: prevent an infinite loop in voidsrc

JDarnley:
This change addresses the issue I was experiencing but I am not sure how "correct" it is.  Please comment if you think there's a better way to handle this.


JDarnley:
This fix appears to be incomplete.  When the clock starts returning a valid value, now might be quite a bit larger than pts leading to many iterations in the loop.  I'm not sure how to handle that though.  voidsrc might need to remember if the past now was UINT64_MAX.

quarium:
You must reset the timer to the interval and leave when now == UINT64_MAX

JDarnley:
@quarium I am not quite sure what you mean.  I now have voidsrc waiting using interval.  It also now doesn't increase pts until it is set to a good value.  This, as a whole, corrects the issues I was encountering with my clock and sink.

cmassiot:
Do you still need this?

JDarnley:
Yes.  The clock still returns UINT64_MAX until the sink pipe gets a flow_def.  We have these changes in a local branch so if you don't want these changes in upstream you can just close this.

PR #599, opened 2019-04-26
upipe-modules: make upipe_sync catch UINT64_MAX values from uclock_now

JDarnley:
Do not drop frames when that happens.  Do not try to wait using that
time.  Do not try to adjust PTS using that time.

Perhaps there is a better way to address this.  However this change has let me get output when using my clock and sink.

I have a cosmetic question.  What is Upipe's opinion on conditional ternary operators?  The if-else expressions could be shortened to `(now == UINT64_MAX) ? ticks_per_frame : pts - now`

### DIFF
--- a/lib/upipe-modules/upipe_sync.c
+++ b/lib/upipe-modules/upipe_sync.c
@@ -799,11 +799,18 @@ static void cb(struct upump *upump)
 
     /* schedule next pic */
     now = uclock_now(upipe_sync->uclock);
-    while (now > upipe_sync->pts) {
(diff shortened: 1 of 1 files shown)


kierank:
IMO you shouldn't run the timer at all if now == UINT64_MAX but not sure what consequences that has.

kierank:
Oh actually timers use the PC clock so it's probably ok

JDarnley:
I did try not using the timers in this but I think I got no output from the pipe.  From what I can tell the output function is only called in the cb function.

PR #600, opened 2019-05-06
SMPTE 2038 Ancillary in TS

funman:


lib/upipe-modules/upipe_vanc_decoder.c
```diff
@@ -0,0 +1,325 @@
...
+            uref_free(pic);
+            continue;
+        }
+
+        upipe_dbg_va(upipe, "y=%d line %d off %d | DID 0x%.2x SDID 0x%.2x DC 0x%.2x",
+                c_not_y, line, offset,
+                s291_get_did(data), s291_get_sdid(data), s291_get_dc(data));
+
+        if (c_not_y)
+            offset = 2 * offset + 1;
```
funman: Wrong as C comes before Y and it's interleaved so we need another bit since we don't know if this is SD or HD

cmassiot:
LGTM

lib/upipe-modules/upipe_vanc_decoder.c
```diff
@@ -0,0 +1,325 @@
...
+
+        data[0] = S291_ADF1;
+        data[1] = S291_ADF2;
+        data[2] = S291_ADF3;
+        data[3] = did;
+        data[4] = sdid;
+        data[5] = dc;
+
+        for (int i = 0; i < (dc & 0xff) + 1; i++) {
+            if (cached_bits == 0) {
```
kierank: ubits might be easier to follow?

kierank: maybe just turn this into ubits_read?

funman: It would not work as initial read is not byte aligned

kierank: You could read the entire bitstream using ubits_read

funman:
Also need to fix offset to start from SAV not EAV, and handle HANC like https://tools.ietf.org/html/rfc8331

funman:
Missing timestamps

lib/upipe-modules/upipe_vanc_decoder.c
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    while (end >= 9) {
+        if (r[0] & 0xfc) { /* 000000 */
+            goto ret;
+        }
+
+        bool c_not_y = r[0] & 0x2;
+
+        unsigned line = ((r[0] & 1) << 10) | (r[1] << 2) | (r[2] >> 6);
```
cmassiot: Wouldn't it be more readable with bitstream helpers?

cmassiot:
Can you fix the conflict ?

PR #604, opened 2019-05-28
ubits: add ubits_get, modify ubits_init

funman:


kierank:
LGTM

cmassiot:
I am not enthusiastic at the idea of adding a true/false argument to ubits_init. Nobody understands by reading the code what the arguments means, you have to go back to the document. I'd be more in favour of an enum argument, or two functions ubits_init_read and ubits_init_write.

PR #607, opened 2019-06-13
speexdsp improvements

kierank:


lib/upipe-speexdsp/upipe_speexdsp.c
```diff
@@ -187,8 +187,14 @@ static bool upipe_speexdsp_handle(struct upipe *upipe, struct uref *uref,
     if (err) {
         ubuf_free(ubuf);
     } else {
+        uint64_t pts_sys = UINT64_MAX;
         ubuf_sound_resize(ubuf, 0, out_len);
         uref_attach_ubuf(uref, ubuf);
+
+        /* XXX: Change pts_prog? */
```
cmassiot: Yes you should change pts_prog also.

kierank: ok

PR #608, opened 2019-06-19
Luajit: add some iterator functions

JDarnley:
Please comment. On the udict commit in particular because it is rather ugly.

The plane iterators are basically copied from the same foreach macros in the C headers.  The udict iterator is inspired by (or based on) udict_dump but without its switch statement.

### DIFF
--- a/luajit/upipe.lua
+++ b/luajit/upipe.lua
@@ -358,8 +358,30 @@ ffi.metatype("struct upipe", {
     end,
 })
 
+--[[
+-- A function to iterate over all the planes of a uref.  For each plane it
+-- returns the plane string (chroma for pic, channel for sound).
+--]]
+local function foreach_plane(func)
+    return function(ref)
+        local plane = ffi.new("const uint8_t *[1]")
+        return function()
+            if C.ubase_check(func(ref, plane)) and plane[0] ~= nil then
+                return ffi.string(plane[0])
+            else
+                return nil
+            end
+        end
+    end
+end
+
 ffi.metatype("struct uref", {
     __index = function (_, key)
+        if key == "pic_foreach_plane" then
+            return foreach_plane(C.uref_pic_iterate_plane)
+        elseif key == "sound_foreach_plane" then
+            return foreach_plane(C.uref_sound_iterate_plane)
+        end
         local f = C[fmt("uref_%s", key)]
         return getter(uref_getters, f, key) or f
     end
--- a/luajit/upipe.lua
+++ b/luajit/upipe.lua
@@ -460,6 +460,99 @@ ffi.metatype("struct uref_mgr", {
     end
 })
 
+local udict_getters = {}
+local function udict_getter(udict_type, ctype, number)
+    local cfunc = C[string.format("udict_get_%s", udict_type)]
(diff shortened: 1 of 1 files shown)


luajit/upipe.lua
```diff
@@ -365,6 +365,28 @@ ffi.metatype("struct uref", {
...
+--
+-- The value for the plane will be invalid on the next iteration so make a copy
+-- if it needs to be kept.
+--]]
+local function foreach_plane(func)
+    return function(ref)
+        local plane = ffi.new("const uint8_t *[1]")
+        return function()
+            if C.ubase_check(func(ref, plane)) and plane[0] ~= nil then
+                return plane[0]
```
nto: Wouldn't it be more useful to return a Lua string here with `ffi.string`? Also, I believe the value is still valid on the next iteration.

luajit/upipe.lua
```diff
@@ -365,6 +365,28 @@ ffi.metatype("struct uref", {
...
+            else
+                return nil
+            end
+        end
+    end
+end
+uref_pic_foreach_plane = foreach_plane(C.uref_pic_iterate_plane)
+uref_sound_foreach_plane = foreach_plane(C.uref_sound_iterate_plane)
+
```
nto: I try to avoid global functions. What about including them in the metatypes instead:

```diff
--- a/luajit/upipe.lua
+++ b/luajit/upipe.lua
@@ -359,7 +359,10 @@ ffi.metatype("struct upipe", {
 })
 
 ffi.metatype("struct uref", {
-    __index = function (_, key)
+    __index = function (ref, key)
+        if key == 'pic_foreach_plane' then
+            return foreach_plane(ref, C.uref_pic_iterate_plane)
+        end
         local f = C[fmt("uref_%s", key)]
         return getter(uref_getters, f, key) or f
     end
```

luajit/upipe.lua
JDarnley: You're probably right.  I didn't want Luajit to convert back and forth since I call other upipe functions using the plane.  Probably a premature (micro)optimisation.

As for being invalid, that is probably the wrong term.  The pointer will change to point to the next plane string whereas Lua strings are immutable.  `ffi.string` is probably best.

luajit/upipe.lua
JDarnley: That would make it more Lua-like

I will try both changes and make additional commits if they work.

JDarnley:
The commit I just pushed works in my program.  Is that how you prefer it?  Maybe I should try adding a metatype for udicts too; see if it works when I put the foreach_attribute in it.

nto:
Yes, it would be great to add the metatype for udict as well.

JDarnley:
How about these 2 commits?

The udict_foreach_attribute function is now local but accessible through the udict metatype.  I thought moving the body of that function into the metatype would be too ugly, especially if the meta type gets expanded in the future.

I also converted some things to Lua types following the suggestion you made on the previous patch.

nto:
Looks fine 👍
Thanks for this contribution!

cmassiot:
Can you squash?

JDarnley:
Done

PR #631, opened 2019-10-01
audio_merge: Make audio channel mismatch message verbose

kierank:
It's not an error, there are use-cases for this

### DIFF
--- a/lib/upipe-modules/upipe_audio_merge.c
+++ b/lib/upipe-modules/upipe_audio_merge.c
@@ -356,7 +356,7 @@ static void upipe_audio_merge_produce_output(struct upipe *upipe, struct upump *
     uint8_t output_channels = 0;
     UBASE_ERROR(upipe, uref_sound_flow_get_channels(upipe_audio_merge->flow_def, &output_channels));
     if (input_channels != output_channels)
-        upipe_err_va(upipe, "total input channels (%"PRIu64") != output flow def (%d), some will be skipped or blanked!",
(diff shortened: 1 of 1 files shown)


lib/upipe-modules/upipe_audio_merge.c
```diff
@@ -356,7 +356,7 @@ static void upipe_audio_merge_produce_output(struct upipe *upipe, struct upump *
     uint8_t output_channels = 0;
     UBASE_ERROR(upipe, uref_sound_flow_get_channels(upipe_audio_merge->flow_def, &output_channels));
     if (input_channels != output_channels)
-        upipe_err_va(upipe, "total input channels (%"PRIu64") != output flow def (%d), some will be skipped or blanked!",
+        upipe_verbose_va(upipe, "total input channels (%"PRIu64") != output flow def (%d), some will be skipped or blanked!",
```
cmassiot: Don't you think it would be better to print it once in set_flow_def ?

kierank: Actually the check is wrong to begin with, I think the intention is to make sure the sum(uref_channels) = sum(flow_def_channels)

JDarnley:
Since I am working on the pipe I took a quick look at this.  AFAICS if you do it in set_flow_def for the subpipes it will print one message for each input until the sum equals the output flow_def, then nothing when equal, then again when greater.

Maybe it is supposed to check that the input urefs match what the input flow_defs say and print a message when they are different indicating some error.

I'm not sure myself so I have just silenced it while I am testing.

cmassiot:
Message was subsequently removed.

PR #634, opened 2019-10-15
pes data alignment indicator fix

nriviera:
pes data alignment indicator is always true but pes_packet_leght=0 and payload in next packet


cmassiot:
I don't get it, with this patch pes_data_alignment will always be false, what is the intent?

nriviera:
Sorry about that, I was misleaded by ts analizer that is issuing a problem related to the pes data alignment in packet that it is in fact caused by a problem parsing the first NAL of the h264 stream.
The code is indeed correct in the way it handles the splicing.
 
By the way I believed that pull request were mediated and would have not triggered a build, I should have asked before

PR #641, opened 2019-11-08
extract_pic: add filter format

quarium:
Add filter format to convert unsupported formats.


cmassiot:
Can you merge with the use of sws and deint just above?

quarium:
Done

PR #646, opened 2019-11-28
upipe-modules/rtp_pcm_pack: add user options to control output size

JDarnley:
Not tested here yet, soon.


JDarnley:
The option to split by time works well for us.  It correctly turns 125us into 6 samples.

We saw that pts_sys needed adjusting on the output urefs for our use so I added that to this PR.  I can separate if you prefer.

JDarnley:
Okay.  I see this stuff now.  I'll look at it all.

PR #649, opened 2019-12-16
upipe-modules/udp: add option to set multicast interface name, miface

JDarnley:
I am not sure on where in the function this option should be set but it works for us in a brief test.


nto:
You can use a shorter prefix in the commit message:
`upipe_udp: add option ...`

kierank:
It's not a multicast interface name, it's the multicast interface address

PR #655, opened 2020-01-09
Add avfilter pipe

quarium:


nto:
Please use a meaningful pull request title, and squash the "fix" commits.

kierank:
How do you plan to deal with latency in avfilter?

nto:
Since latency depends on the caller-provided filtergraph, I guess the latency attribute needs to be manually set by the caller as well? I don't know of any automated way to do it unfortunately.

nto:
> Please use a meaningful pull request title, and squash the "fix" commits.

Also, commented code should probably be removed.

include/upipe/uref_pic_flow_formats.h
```diff
@@ -608,6 +657,17 @@ static const struct uref_pic_flow_format uref_pic_flow_format_rgba64be = {
 
 UREF_PIC_FLOW_FORMAT_HELPER(rgba64be);
 
+/** @This is the description of the rgba64le. */
```
nto: missing "format" at the end of description.

JDarnley:
I have cherry picked these commits (and one other needed to build) into a private branch and taken them for a spin.  A bit of quick hacking in a test program I can get it to work reasonably well as a video source.

I have an issue but it might be explained by missing commits from Upipe master or because I am not building the pipeline correctly.

The video frames are generated far too quickly.  The program has a real time video output and a hardware clock to go along with it.  Previously video frames would come from zoneplate source (with the real source being void source in that pipe) and it would generate videos at exactly the rate needed.  This is generating video frames at up to 240fps.

Thank you.  It has been pleasant to look at something other than zoneplate.

lib/upipe-av/upipe_avfilter.c
```diff
@@ -0,0 +1,1637 @@
...
+    /** sentinel */
+    UPROBE_AVFILT_SUB_SENTINEL = UPROBE_LOCAL,
+    /** the filter was updated (void) */
+    UPROBE_AVFILT_SUB_UPDATE,
+};
+
+/** @internal @This enumerates the supported media types. */
+enum upipe_avfilt_sub_media_type {
+    /** audio */
+    UPIPE_AVFILT_SUB_MEDIA_TYPE_VIDEO,
```
nto: Inverted comments?

PR #659, opened 2020-02-17
upipe_speexdsp: check that the ubuf_mgr exists before using it

JDarnley:
Drop input urefs if it doesn't exist.

On IRC I asked about the following problem: "I need to deport upipe_speexdsp to a worker thread but it doesn't get a ubuf_mgr before trying to create a ubuf with it which causes an assert.  What is the preferred way to have the pipe check whether it is ready to handle input urefs?"

This is what I quickly put together for our usage because I wasn't sure how else to handle it.

### DIFF
--- a/lib/upipe-speexdsp/upipe_speexdsp.c
+++ b/lib/upipe-speexdsp/upipe_speexdsp.c
@@ -126,6 +126,11 @@ static bool upipe_speexdsp_handle(struct upipe *upipe, struct uref *uref,
 {
     struct upipe_speexdsp *upipe_speexdsp = upipe_speexdsp_from_upipe(upipe);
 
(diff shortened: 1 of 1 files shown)


cmassiot:
Well the correct way would be to queue the incoming urefs, and process them when you get a ubuf manager. For a simple example look at upipe_genaux.c.

JDarnley:
I've finally come back to this PR.

Is all I need to do just return false?  This is already in the handle function, it already uses HELPER_INPUT.

cmassiot:
Yes I think so.

nto:
Can you squash the commits into a single one?

JDarnley:
done and force pushed

PR #676, opened 2020-04-20
upipe-av: add an AVFrame ubuf manager

quarium:


nto:
Please use a better pull request title.

PR #679, opened 2020-04-27
Add pixel formats

quarium:


cmassiot:
I would have loved some details in the comments on what those new formats are, or pointers to some documentation.

PR #682, opened 2020-05-04
upipe_pthread: support custom niceness value

quarium:


kierank:
My understanding is that setpriority will change the niceness for all threads of the process. Is that intentional?

kierank:
Ok apparently this is a bug in Linux (see bottom):

https://linux.die.net/man/2/setpriority

quarium:
> Ok apparently this is a bug in Linux (see bottom):
> 
> https://linux.die.net/man/2/setpriority

Yes this is intentional =)

PR #690, opened 2020-05-13
ubuf_av: fix leaks on allocation failure

quarium:
Depend on nto pull request

### DIFF
--- a/lib/upipe-av/ubuf_av.c
+++ b/lib/upipe-av/ubuf_av.c
@@ -134,7 +134,6 @@ static void ubuf_av_free(struct ubuf *ubuf)
             free(buf[i]);
         free(buf);
     }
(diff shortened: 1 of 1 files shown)


lib/upipe-av/ubuf_av.c
```diff
@@ -134,8 +134,10 @@ static void ubuf_av_free(struct ubuf *ubuf)
             free(buf[i]);
         free(buf);
     }
-    av_frame_unref(ubuf_av->frame);
-    av_frame_free(&ubuf_av->frame);
+    if (ubuf_av->frame) {
+        av_frame_unref(ubuf_av->frame);
```
nto, line 138: I believe this `av_frame_unref()` is not required since `av_frame_free()` already unrefs the frame?

PR #697, opened 2020-06-06
rtp_feedback: Check RTCP SR packet is large enough

kierank:


kierank:
Seems like it's needed but I wasn't able to easily cause a crash (possibly because of memory pools)

PR #710, opened 2020-07-15
upipe_audio_merge: upstreaming some improvements

JDarnley:
We have some more things to come, such as interleaved support, but that might depend on another attribute being added first.

### DIFF
--- a/lib/upipe-modules/upipe_audio_merge.c
+++ b/lib/upipe-modules/upipe_audio_merge.c
@@ -524,8 +524,12 @@ static struct upipe *upipe_audio_merge_alloc(struct upipe_mgr *mgr,
 static int upipe_audio_merge_check(struct upipe *upipe, struct uref *flow_format)
 {
     struct upipe_audio_merge *upipe_audio_merge = upipe_audio_merge_from_upipe(upipe);
-    if (flow_format != NULL)
+    if (flow_format != NULL) {
+        // FIXME: why do we need to do this?!
+        if (!ubase_check(uref_clock_set_latency(flow_format, upipe_audio_merge->latency)))
+            upipe_throw_error(upipe, UBASE_ERR_ALLOC);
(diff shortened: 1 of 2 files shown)


lib/upipe-modules/upipe_audio_merge.c
```diff
@@ -210,6 +227,11 @@ static struct upipe *upipe_audio_merge_sub_alloc(struct upipe_mgr *mgr,
 
     struct upipe_audio_merge_sub *upipe_audio_merge_sub =
                             upipe_audio_merge_sub_from_upipe(upipe);
+    struct upipe_audio_merge *upipe_audio_merge =
+        upipe_audio_merge_from_sub_mgr(mgr);
+
+    /* Is this the first subpipe to be allocated? */
+    upipe_audio_merge_sub->first_subpipe = ulist_depth(&upipe_audio_merge->inputs) == 0;
```
cmassiot, line 263: What if the first subpipe is released?

JDarnley: I am not sure.  I think it would lock the pipe from making any changes until all inputs are released.  I think I never encountered that because our pipeline changes involves all inputs being released.

Do you want me to try something else?

cmassiot: For this kind of pipe where an input is "privileged" I would have done it differently, for instance have the privileged input send data to the superpipe, and only have non-privileged inputs as subpipes. I don't know if it's feasible in your case.

JDarnley: I have an idea.  If we are really releasing all inputs on the change then I should make the pipe clear the stored flow_def when the last pipe is released.  I'll experiment.

JDarnley: I can no longer reproduce the issue we had for which I wrote this.  I will probably drop that commit when I rebase to add the rest of the commits.

PR #711, opened 2020-07-16
upipe_grid: fix warnings

quarium:


lib/upipe-modules/upipe_grid.c
```diff
@@ -387,7 +387,8 @@ static void upipe_grid_in_update_pts(struct upipe *upipe, uint64_t next_pts)
                 uref_free(uref);
                 continue;
             }
-            upipe_warn(upipe, "keeping last input");
+            upipe_warn_va(upipe, "keeping last input %.3f ms",
+                          1000. * (next_pts - rebase_pts) / UCLOCK_FREQ);
```
cmassiot, line 391: Is there any risk this might flood a syslog?

quarium: Yes it does, I'll do it differently

PR #727, opened 2020-08-19
ubuf_pic: improve ubuf_pic_clear

nto:
Use ubuf_pic_plane_set_color with a chroma-specific color pattern.
Improve speed of ubuf_pic_plane_set_color for patterns larger than 1
byte by copying whole lines instead of a few bytes at a time, leading
to a 12x speed increase for yuv420p10le.


kierank:
Looks ok. Makes sense as it is very slow for us as well sometimes

quarium (approved):

PR #728, opened 2020-08-24
Add uref/udict dump functions

quarium:


tests/Makefile.am
```diff
@@ -134,6 +135,7 @@ TESTS = \
 	uprobe_uref_mgr_test \
 	uref_std_test \
 	uref_uri_test.sh \
+	uref_dump_test \
```
cmassiot, line 138: Shouldn't we compare the output with the saved output?

PR #729, opened 2020-08-24
upipe: add ulist last element getter

quarium:


include/upipe/ulist.h
```diff
@@ -168,6 +169,19 @@ static inline struct uchain *ulist_peek(struct uchain *ulist)
     return ulist->next;
 }
 
+/** @This returns a pointer to the last element of the list (without
+ * removing it).
+ *
+ * @param ulist pointer to a ulist
+ * @return pointer to the last element or NULL if the list is empty
+ */
+static inline struct uchain *ulist_last(struct uchain *ulist)
```
cmassiot, line 178: Perhaps ulist_peek_last? I'd like to keep in the name of the function the idea that it is just a "peek".

PR #731, opened 2020-08-24
uprobe_stdio: add time output

quarium:


lib/upipe/uprobe_stdio.c
```diff
@@ -171,6 +179,18 @@ static int uprobe_stdio_throw(struct uprobe *uprobe, struct upipe *upipe,
...
+        if (uprobe_stdio->time_format) {
+            time_t t = time(NULL);
+            struct tm tm;
+            struct tm *tm_p = localtime_r(&t, &tm);
+            char str[TIME_FORMAT_MAX_SIZE];
+            const char *sfx = " ";
+            if (!strftime(str, sizeof (str), uprobe_stdio->time_format, tm_p)) {
+                str[0] = '\0';
+                sfx = "";
+            }
```
nto, line 191: This code should be hoisted outside the loop, because all lines of a multi-line log should have the same timestamp.

nto (changes requested):

lib/upipe/uprobe_stdio.c
```diff
@@ -238,6 +262,18 @@ void uprobe_stdio_set_color(struct uprobe *uprobe, bool enabled)
...
+ * @param uprobe pointer to probe
+ * @param format strftime format or NULL to disable
+ */
+void uprobe_stdio_set_time_format(struct uprobe *uprobe, const char *format)
+{
+    struct uprobe_stdio *uprobe_stdio = uprobe_stdio_from_uprobe(uprobe);
+    free(uprobe_stdio->time_format);
+    uprobe_stdio->time_format = format ? strdup(format) : NULL;
+}
+
```
nto, line 276: Perhaps it would be more appropriate to return `UBASE_ERR_ALLOC` if `strdup` fails?

nto (changes requested):

nto (approved):

PR #733, opened 2020-08-24
upipe_bmd_sink: support external clock input

quarium:


kierank:
Does this actually work?

lib/upipe-blackmagic/upipe_blackmagic_sink.cpp
```diff
@@ -376,14 +376,32 @@ class callback : public IDeckLinkVideoOutputCallback
...
+            case 2:
+            case 3:
+                log = upipe_warn_va;
+                break;
+            default:
+                result = 4;
+                log = upipe_err_va;
+                break;
+        }
+        log(&upipe_bmd_sink->upipe, "%p Frame %s", frame, Result_str[result]);
```
nto, line 407: Use `upipe_log_va` instead of this function pointer.

nto (changes requested):

PR #745, opened 2020-09-02
upipe_s302_framer: add low latency output

JDarnley:
Quite a hack with lots of duplicated code and no option to control it.


JDarnley:
Does anyone have an idea or a suggestion how I could pass a run time option for enabling this to the pipe framer through the ts demux and the auto framer?  For our use it doesn't have to change dynamically but it would be better to have something other than at compile time.

JDarnley:
Is this any good for a run-time option to enable it?  Rather circuitous but it works.

PR #749, opened 2020-10-30
upipe-modules: add upipe_pad

JDarnley:


lib/upipe-modules/upipe_pad.c
```diff
@@ -0,0 +1,296 @@
...
+            return UBASE_ERR_UNHANDLED;
+
+        case UPIPE_SET_FLOW_DEF: {
+            struct uref *flow = va_arg(args, struct uref *);
+            return upipe_pad_set_flow_def(upipe, flow);
+        }
+    }
+}
+
+/** @internal @This is the static zoneplate source pipe manager. */
```
cmassiot, line 280: Not zoneplate.

tests/upipe_pad_test.c
```diff
@@ -0,0 +1,320 @@
+/*
+ * Copyright (C) 2014-2016 OpenHeadend S.A.R.L.
+ *
+ * Authors: Sebastien Gougelet
+ *          Christophe Massiot
```
cmassiot, line 5: Are you sure about the names ?

cmassiot (changes requested):
It is sad that you do a memcpy of the picture which seems unnecessary. Why don't you just set hmprepend, hmappend, vprepend and vappend during a flow format negotiation?

JDarnley:
Oh.  Looks like I copy-pasted too much.  Sorry and thank you for catching all that.

I need to copy because there seems to be no opportunity to modify the ubuf_mgr request that comes out of upipe_avcodec_decode (avcdec) when it is deported to a worker thread.  avcdec calls its demand_ubuf_mgr function from UPIPE_HELPER_UBUF_MGR so it falls back to getting one from the probe when the request isn't answered in the normal manner.  I assume it doesn't pass the thread boundary so it can't pass to the usual pipe we use, upipe_ntsc_prepend.

This is only something we recently discovered.  Usually we have upipe_ffmt and upipe_sws between the two to convert to 4:2:2 but with that as input they don't need to do anything so they don't send their own asynchronous request for a ubuf_mgr which upipe_ntsc_prepend normally modifies.

include/upipe-modules/upipe_pad.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * Pad picture urefs with blank space.
+ *
+ * Copyright (C) 2020 Open Broadcast Systems Ltd.
+ *
+ * Authors: James Darnley
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
```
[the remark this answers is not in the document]
JDarnley: Changed to MIT

include/upipe-modules/upipe_pad.h
```diff
@@ -0,0 +1,48 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+/** @file
+ * @short Upipe module to pad picture urefs with blank space.
+ */
+
+#ifndef _UPIPE_FILTERS_UPIPE_PAD_H_
```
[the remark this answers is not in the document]
JDarnley: Changed

lib/upipe-modules/upipe_pad.c
JDarnley: Corrected

tests/upipe_pad_test.c
JDarnley: Corrected

cmassiot:
Could you rebase the pull request?

PR #753, opened 2020-11-10
upipe_mpgaf: compute octetrate after encapsulation

quarium:
Compute octetrate based on the output packet duration and size after
encapsulation for AAC streams so they take account of the possible
encapsulation overhead.
Use parsed octetrate for MPEG layers 1, 2 and 3 as the encapsulation
doesn't change.

### DIFF
--- a/lib/upipe-framers/upipe_mpga_framer.c
+++ b/lib/upipe-framers/upipe_mpga_framer.c
@@ -519,15 +519,8 @@ static bool upipe_mpgaf_parse_adts(struct upipe *upipe)
         asc_aot = ASC_TYPE_SBR;
     }
 
-    /* Calculate octetrate assuming the stream is CBR. */
-    uint64_t octetrate = adts_length * samplerate / samples;
-    /* Round up to a multiple of 8 kbits/s. */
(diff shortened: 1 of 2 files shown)


lib/upipe-framers/upipe_mpga_framer.c
```diff
@@ -1762,6 +1768,46 @@ static bool upipe_mpgaf_work_raw(struct upipe *upipe, struct uref *uref,
...
+                                      struct uref *flow_def)
+{
+    struct upipe_mpgaf *upipe_mpgaf = upipe_mpgaf_from_upipe(upipe);
+    const char *def = NULL;
+
+    UBASE_RETURN(uref_flow_get_def(flow_def, &def));
+    upipe_mpgaf->input_latency = 0;
+    uref_clock_get_latency(flow_def, &upipe_mpgaf->input_latency);
+    upipe_mpgaf->encaps_input = uref_mpga_flow_infer_encaps(flow_def);
+    upipe_mpgaf->octetrate = 0;
```
quarium, line 1787: Is it right?

PR #763, opened 2020-12-01
upipe_ts_scte: check length

quarium:


lib/upipe-ts/upipe_ts_scte104_decoder.c
```diff
@@ -609,9 +644,24 @@ upipe_ts_scte104d_insert_segmentation_descriptor(struct upipe *upipe,
...
     if (device_restrictions)
         uref_ts_scte35_desc_seg_set_device(desc, device_restrictions);
-    if (insert_sub_info) {
-        uref_ts_scte35_desc_seg_set_sub_num(desc, sub_num);
-        uref_ts_scte35_desc_seg_set_sub_expected(desc, sub_expected);
+    switch (type_id) {
+        case 0x34:
+        case 0x36:
+            if (!insert_sub_info) {
+                upipe_warn(upipe, "no sub info flag when needed");
```
cmassiot, line 651: We don't need the warning: it is legal to have pre-2016 SCTE 104 without this structure.

cmassiot (changes requested):

lib/upipe-ts/upipe_ts_scte104_decoder.c
quarium: It won’t be when I’ll add the data length check.

PR #767, opened 2020-12-07
upipe_sync: Use correct latency based on framerate

kierank:


nto:
Perhaps the commit message should point out this is about the `upipe_sync` pipe?

kierank:
Yes, that's true

PR #790, opened 2021-03-24
upipe_rtp_fec: add latency getter

nto:


kierank:
Bear in mind we add the latency to date_sys in the pipe

PR #791, opened 2021-03-24
upipe_rtp_fec: support matrix size bigger than 128

nto:


kierank:
Ok if tested

nto:
Working fine with a 20x20 matrix.

PR #797, opened 2021-04-21
upipe_h264f: add support for separated fields

quarium:


cmassiot:
Perhaps add "for length-based formats" (mp4).

PR #801, opened 2021-04-23
upipe: fix sub manager initialization

quarium:
Make sure all the sub pipe manager structure is initialized to prevent
use of uninitialized memory.


cmassiot:
LGTM.

PR #806, opened 2021-06-19
upipe_blackmagic: Support segmented teletext ubufs

kierank:
Fixes high bitrate teletext streams


lib/upipe-blackmagic/upipe_blackmagic_sink.cpp
```diff
@@ -269,6 +269,9 @@ struct upipe_bmd_sink {
     // XXX: should counter be per-field?
     uint16_t op47_sequence_counter[2];
 
+    /** OP47 teletext buffer. 10 is the max number of units */
```
cmassiot, line 272: Then I'd rather you #define MAX_TELX_UNITS 10 somewhere.

PR #846, opened 2021-12-21
upipe_queue_source: fix race condition

quarium:
Use a second refcount for the pumps as they may be used in another
thread (spoiler: it's meant to be).
Without this refcount, the pumps may use then release the external
refcount when dispatching (see upump_common_dispatch) after it has
reached zero in another thread.
Because urefcount_release is not thread safe in this case, there is 3
possible outcomes:
    - nothing wrong because the refcount callback was already set to
    NULL by the other thread (a.k.a. the planets are aligned)
    - urefcount_release SEGV because refcount->cb was set to NULL by the
    other thread after it was tested
    - the refcount callback is called twice

The first problem here is that urefcounts are NOT thread safe if it's
allowed to use a reference which was already released. But it seems
there are some pipes which are designed to throw events and support
control commands when released (upipe_ts_demux which throws split update
when released and handle split_iterate for instance) making them
dangerous to use in separate threads.

The second problem is that the pumps doesn't use the refcount that is
given to them during all their lifetime but only when dispatching making
it useless with multiple threads.

This second refcount bypasses the two problems by using another refcount
which is used during the whole lifetime of the pumps.

### DIFF
--- a/lib/upipe-modules/upipe_queue_source.c
+++ b/lib/upipe-modules/upipe_queue_source.c
@@ -41,6 +41,7 @@
 #include "upipe/upump.h"
 #include "upipe/upipe_helper_upipe.h"
 #include "upipe/upipe_helper_urefcount.h"
+#include "upipe/upipe_helper_urefcount_real.h"
 #include "upipe/upipe_helper_output.h"
 #include "upipe/upipe_helper_upump_mgr.h"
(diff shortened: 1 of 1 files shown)


quarium:
Maybe we should forbid to use a released refcount?

PR #868, opened 2022-07-25
more optimization for ubuf_pic and upipe_set_color

JDarnley:
The last commit, the benchmark, isn't supposed to get committed.


JDarnley:
The performance is made worse on master so I'm closing this.  Honestly WTF

JDarnley:
My own stupid fault.  I was building in a directory I had previously configured with -O0.

The main patch increases the benchmark from ~6500 calls/s to ~7000.

PR #874, opened 2022-10-14
upipe_avcodec_decode: use pic.sub. for subtitles

quarium:


JDarnley:
I would need to test our software linked with master to be sure but I think this will now cause us any problems.  We use ubase_ncmp to check the flows are "pic." in a few places but that should still return true with this change.

PR #901, opened 2023-06-14
Handle NULL pointer in _from_/_to_ helpers

quarium:


cmassiot:
These macros are used absolutely everywhere all the time, I am a bit worried of the impact on performance. Would ubase_assert() work?

PR #909, opened 2023-06-19
upipe_filter_format: pass through the flow request

quarium:
Override the wanted attributes and pass through the flow format request
to the downstream pipe. This lets the downstream pipe add some useful
attributes for the upstream pipe.


quarium:
@cmassiot

cmassiot (approved):

PR #911, opened 2023-06-23
ubuf_pic: assume alpha to be always fullrange

quarium:


lib/upipe/ubuf_pic.c
```diff
@@ -62,9 +62,23 @@ int ubuf_pic_plane_clear(struct ubuf *ubuf, const char *chroma,
...
+        } else {
+            SET_COLOR(16, 16, 16, 0);
+        }
+    } else if (MATCH("a8r8g8b8") || MATCH("a8b8g8r8")) {
+        if (fullrange) {
+            SET_COLOR(0, 0, 0, 0);
+        } else {
+            SET_COLOR(0, 16, 16, 16);
+        }
+    } else if (MATCH("y8") || MATCH("y16l") || MATCH("y16b") ||
```
nto, line 80: Is it the correct color for `y16l` and `y16b`?

nto (changes requested):

lib/upipe/ubuf_pic.c
quarium: Probably not but I didn't change that.
@kierank: any idea?

nto (approved):

PR #926, opened 2023-08-03
Add ID3v2 decaps/encaps pipes

quarium:


include/upipe-modules/upipe_id3v2_decaps.h
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2023 EasyTools
+ *
```
nto, line 3: Missing authorship information.

PR #983, opened 2024-04-06
udp_source/sink: Check for exiting fd and close it during set_fd

kierank:


cmassiot:
I think you mean “existing”.

PR #991, opened 2024-06-21
s12m timecodes

JDarnley:
Supersedes #530 

Comments very much welcome

### DIFF
--- a/include/upipe/uref_attr_s12m.h
+++ b/include/upipe/uref_attr_s12m.h
@@ -1,0 +1,184 @@
+/*
+ * Copyright (C) 2024 Open Broadcast Systems Ltd
+ *
+ * Authors: James Darnley
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject
+ * to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/** @file
+ * @short Upipe uref s12m attributes handling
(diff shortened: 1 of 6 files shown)


kierank:
I'm not sure what I think about storing them in a uref as BCD. I see why you did it but it's a bit ugly. But then again anything else would add complexity.

include/upipe/uref_attr_s12m.h
```diff
@@ -0,0 +1,184 @@
...
+#include <stdint.h>
+
+/** @this returns a uint32_t read from possibly unaligned data
+ *
+ * Timecode packs and the count are native endian values.
+ */
+static inline uint32_t uref_attr_s12m_read(const uint8_t *data)
+{
+    /* TODO: big endian */
+    return (uint32_t)data[0] | data[1] << 8 | data[2] << 16 | data[3] <<24;
```
kierank, line 52: cosmetics

nto:
Perhaps the functions in `uref_attr_s12m.h` should validate their input values and return `UBASE_ERR_INVALID` when out of range?

JDarnley:
I added that, well something that a user can choose to use if they don't trust the data.  I assume you wanted the check in the to/from functions and for those to return an error code rather than the timecode pack itself.

I wonder if I should change the check function to return false if the count is 0.  Is something, someone going to add a 4 byte attribute just to hold one 0?

I also needed a function to set a flag in a different position for NTSC and PAL so that got created.

lib/upipe-av/upipe_avcodec_decode.c
```diff
@@ -1324,6 +1324,13 @@ static void upipe_avcdec_output_pic(struct upipe *upipe, struct upump **upump_p)
     if (side_data)
         uref_pic_set_cea_708(uref, side_data->data, side_data->size);
 
+#ifdef AV_FRAME_DATA_S12M_TIMECODE
```
JDarnley, line 1327: Can anyone explain how this #ifdef can fail with an ffmpeg version new enough to have the side data?

nto: `AV_FRAME_DATA_S12M_TIMECODE` is an enum value, not a preprocessor define. To check for availability, you can use `#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(56, 20, 100)`.

JDarnley: Yes of course!  How dumb of me.  Is a check really needed?  What minimum version is assumed elsewhere?  I'll look for some other checks that already exist and try to judge.

JDarnley: S12M was added in 4.1 and there's a check for libavcodec version available in 4.0 so I will use your check.  Thank you.

lib/upipe-blackmagic/upipe_blackmagic_source.cpp
```diff
@@ -545,6 +545,12 @@ HRESULT DeckLinkCaptureDelegate::VideoInputFrameArrived(
             else if (upipe_bmd_src->tff)
                 uref_pic_set_tff(uref);
 
+            IDeckLinkTimecode *timecode;
+            if (VideoFrame->GetTimecode(bmdTimecodeRP188Any, &timecode) == S_OK) {
+                uint32_t bcd[2] = { 1, timecode->GetBCD() };
```
JDarnley, line 550: I discovered that the BCD from the Decklink API is byte-swapped compared to the ffmpeg side data.  I wonder if I should call a different function (GetComponents) or whether I can just byte-swap it.

Do we use any builtin functions anywhere?  I see some such as "expect" and "ctz" which are guarded by a gnuc check, and some "cpu_supports" guarded by other checks.

I think I'll just call the other function.

JDarnley:
I can't get timecodes out of upipe_blackmagic_sink with any of the equipment I have available.  I'm not sure why.  The driver might be too old or there could be a problem with the C++ code.  I'll investigate some more in the future.

PR #1002, opened 2024-08-16
upipe_opus_framer: support variable frame sample sizes

JDarnley:


lib/upipe-framers/upipe_opus_framer.c
```diff
@@ -241,6 +241,28 @@ static bool upipe_opusf_parse_header(struct upipe *upipe)
     while( size == 0xff );
     upipe_opusf->consume_bytes = idx+2 ;
 
+    uint8_t toc;
+    if (unlikely(!ubase_check(uref_block_extract(upipe_opusf->next_uref,
+                                                 idx + 2, 1, &toc)))) {
+        upipe_err(upipe, "unable to read TOC");
```
kierank, line 247: This shouldn't be an error, its dbg

PR #1034, opened 2024-10-09
upipe-av: add support for hdr10 metadata

nto:


JDarnley:
Rather late but ... For the decode portion of this do you not want to get transfer characteristics and similar attributes from the AVCodecContext too?

nto:
Framers already provide the colorimetry specifications. Those could probably be replaced with values from AVCodecContext, if there is a way to be notified when they are changed?

JDarnley:
Yes I do see those attributes coming from the framer in the flow_def, which is part of my problem.  I see them change in the output of the framer (and the couple of pipes before the decoder).  The new flow_def is accepted by the decoder but it doesn't issue a new one with the changed attributes.

From my limited understanding I think the decoder gathers its own attributes into a uref (sidedata like light level) and checks if these attributes have changed.  If they have changed the pipe issues a new flow_def which is the union(?) between the input and "internal" attributes.  If they have not changed nothing happens (other than a frame is output).  Because the Transfer Characteristics are only in the input flow_def the pipe doesn't see they changed when it does its check so it doesn't update them in the output.

nto:
You are probably right. The decoder should be fixed to properly propagate such changes.

PR #1040, opened 2024-11-08
udp: set fd to -1 if URI is NULL

funman:


nto:
Just use `ubase_clean_fd()` instead of `close()` in _upipe_udp_sink_? _upipe_udp_source_ already does this, so it does not require any change, does it?

funman:
> Just use `ubase_clean_fd()` instead of `close()` in _upipe_udp_sink_? _upipe_udp_source_ already does this, so it does not require any change, does it?

You're right udp_source is already doing it.

Thanks

PR #1065, opened 2025-03-04
upipe_avfilter: factorize main and sub pipe code

quarium:


nto:
Can you add a test case using the null/anull filters to check both simple and complex filtergraph paths?

nto (approved):

PR #1087, opened 2025-07-15
h265_framer improvements, crash fixes and cropping

kierank:


kierank:
I copied it from the H.265 spec.

PR #1164, opened 2026-05-05
upipe_ts_tstd: prevent redundant flow def negotiation

quarium:
Prevent new flow def negotiation when setting the max delay to the currently configured value.


nto (changes requested):
Commit title is misleading. Shouldn't it be `prevent redundant flow def negociation`?

nto (approved):

PR #1166, opened 2026-05-20
upipe_h265_framer: Fix chroma_array_type derivation

kierank:
Fixes vsize on 10-bit 4:2:0 streams


JDarnley:
Is this right?  I'm hardly familiar with the spec Table 6-1 says that the chroma subsampling is 1 for monochrome and 4:4:4 (obviously).  Those modes are indicated with a combination of chroma_format_idc and separate_colour_plane_flag.  Monochrome being 0 and 0 and 4:4:4 being 3 and 0 or 3 and 1.  I'm not sure what chroma_array_type is supposed to be since it is only a local variable.  If separate_colour_plane_flag can only be set for chroma_format_idc=3 then ... maybe?  It is only read from the bitstream when chroma_format_idc=3.  What about fuzzing or malformed streams?

Ultimately I don't know enough.

PR #1168, opened 2026-05-28
upipe_ts_scte35_decoder: fix use-after-unmap in command handlers

kierank:
uref_ts_scte35_set_message reads from the scte35 pointer obtained via uref_block_read. Call it before uref_block_unmap in all three command handlers so the pointer is still valid when the data is copied.


cmassiot:
uref_ts_scte35_set_message does not write to the buffer but to an udict, so it can be done after the unmap.

Issue #1171, opened 2026-06-12
Possible use after free in upipe_ts_demux.c


JDarnley:
I have asan telling me about this error with these trimmed stack traces
```
==664562==ERROR: AddressSanitizer: heap-use-after-free on address 0x5120000a43c8 at pc 0x5555557f3cf5 bp 0x7fffffffc810 sp 0x7fffffffc800
READ of size 1 at 0x5120000a43c8 thread T0
    #0 0x5555557f3cf4 in desc09_get_sysid /home/obe/stress-test-asan/pfx/include/bitstream/mpeg/psi/desc_09.h:58
    #1 0x555555803834 in upipe_ts_demux_program_pmtd_new_flow_def /home/obe/stress-test-asan/upipe/lib/upipe-ts/upipe_ts_demux.c:1597

0x5120000a43c8 is located 264 bytes inside of 286-byte region [0x5120000a42c0,0x5120000a43de)
freed by thread T0 here:
    #0 0x7ffff767fc38 in __interceptor_realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x555555fe67ab in umem_alloc_realloc /home/obe/stress-test-asan/upipe/lib/upipe/umem_alloc.c:77
    #2 0x555555feb30c in umem_realloc ../../include/upipe/umem.h:126
    #3 0x555555fec805 in udict_inline_set /home/obe/stress-test-asan/upipe/lib/upipe/udict_inline.c:453
    #4 0x555555fedad7 in udict_inline_control /home/obe/stress-test-asan/upipe/lib/upipe/udict_inline.c:541
    #5 0x5555557e363d in udict_control_va ../../include/upipe/udict.h:253
    #6 0x5555557e3764 in udict_control ../../include/upipe/udict.h:268
    #7 0x5555557e53fd in udict_set ../../include/upipe/udict.h:563
    #8 0x5555557e5f9c in udict_set_unsigned ../../include/upipe/udict.h:749
    #9 0x5555557ea273 in uref_attr_set_unsigned ../../include/upipe/uref_attr.h:240
    #10 0x5555557f22db in uref_ts_flow_set_capid ../../include/upipe-ts/uref_ts_flow.h:105
    #11 0x555555803807 in upipe_ts_demux_program_pmtd_new_flow_def /home/obe/stress-test-asan/upipe/lib/upipe-ts/upipe_ts_demux.c:1596
```

These line translate to the following lines in master:
[...]

PR #1172, opened 2026-06-12
upipe-ts: fix use-after-free parsing PMT CA descriptors

kierank:
uref_ts_flow_get_descriptor() returns a pointer directly into flow_def's udict inline buffer. The uref_ts_flow_set_capid()/set_sysid() calls (and the cat_onid/cat_esid sets inside upipe_ts_catd_parse_bissca_descs) add attributes to that same uref, which can realloc the udict buffer and invalidate the descriptor pointer. Subsequent reads through it (desc09_get_sysid, the loop advance, and the BISS-CA parse) were use-after-free, as caught by AddressSanitizer at desc09_get_sysid().

Parse a stack copy of the descriptor list instead, so the descriptor reads are independent of any reallocation of flow_def. This mirrors the existing VLA copy pattern in upipe_ts_pmtd_parse_descs(); len is bounded by the PMT program_info_length.

### DIFF
--- a/lib/upipe-ts/upipe_ts_demux.c
+++ b/lib/upipe-ts/upipe_ts_demux.c
@@ -1767,11 +1767,18 @@ static int upipe_ts_demux_program_pmtd_new_flow_def(
 
     upipe_ts_demux_program_build_flow_def(upipe);
 
-    const uint8_t *desc = NULL;
+    const uint8_t *desc_attr = NULL;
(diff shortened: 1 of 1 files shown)


kierank:
Claude verified the VLA can't get > 4096 bytes.

kierank:
Fixes https://github.com/Upipe/upipe/issues/1171

PR #1173, opened 2026-06-12
upipe_mpgv_framer: guard against NULL next_uref when flushing buffere…

kierank:
…d frame

Avoid handling a buffered frame when next_uref is NULL, both during work and on free, preventing a NULL dereference.


kierank:
This fixes a genuine crash

nto (changes requested):
- Please try to follow basic coding style conventions such as the 50/72 Rule for commit messages and max line length for code.
- The test cases should be added to the already existing `upipe_mpgv_framer_test` executable.
- Co-Authored-By trailers should be removed as Upipe's commit history is not the place for advertising proprietary products from AI companies.

PR #1178, opened 2026-06-18
uprobe_dup: new probe that duplicates an event to a second probe

funman:
Usage example:

+    FILE *f = fopen("log.txt", "w");
+    assert(f);
+    struct uprobe *uprobe_log = uprobe_stdio_alloc(NULL, f, loglevel);
+    uprobe_stdio_set_color(uprobe_log, true);
+    assert(uprobe_log != NULL);
+
     uprobe_main = uprobe_stdio_alloc(NULL, stderr, loglevel);
     assert(uprobe_main != NULL);
+    uprobe_main = uprobe_dup_alloc(uprobe_main, uprobe_log, UPROBE_LOG);
+    assert(uprobe_main != NULL);
+    uprobe_release(uprobe_log);

To get both terminal output and file output (with colors)

### DIFF
--- a/lib/upipe/uprobe_dup.c
+++ b/lib/upipe/uprobe_dup.c
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2026 Open Broadcast Systems Ltd
+ *
+ * Authors: Rafaël Carré
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/** @file
+ * @short probe catching events and duplicating them to another probe
+ */
+
+#include "upipe/uprobe.h"
+#include "upipe/uprobe_dup.h"
+#include "upipe/uprobe_helper_alloc.h"
+
+/** @internal @This catches events thrown by pipes.
+ *
+ * @param uprobe pointer to probe
+ * @param upipe pointer to pipe throwing the event
+ * @param event event thrown
+ * @param args optional event-specific parameters
+ * @return an error code
+ */
+static int uprobe_dup_throw(struct uprobe *uprobe, struct upipe *upipe,
+                            int event, va_list args)
+{
+    struct uprobe_dup *uprobe_dup =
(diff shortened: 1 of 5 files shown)


nto (changes requested):
- Can you add a test case?
- Please fix the coding style so that function parameters are correctly aligned
- Commit title should be in imperative style and comply with the _50/72 Rule_ (eg: `uprobe_dup: duplicate an event to a second probe`)
- You may remove error handling from the example and use more helpful variable names (eg: `file_output` and `term_output` for each uprobe_stdio instance)

lib/upipe/uprobe_dup.c
```diff
@@ -0,0 +1,78 @@
...
+/** @internal @This catches events thrown by pipes.
+ *
+ * @param uprobe pointer to probe
+ * @param upipe pointer to pipe throwing the event
+ * @param event event thrown
+ * @param args optional event-specific parameters
+ * @return an error code
+ */
+static int uprobe_dup_throw(struct uprobe *uprobe, struct upipe *upipe,
+                               int event, va_list args)
```
nto:
```suggestion
                            int event, va_list args)
```

tests/uprobe_dup_test.c
```diff
@@ -0,0 +1,95 @@
...
+        case UPROBE_LOCAL:
+            events++;
+            break;
+    }
+    return UBASE_ERR_NONE;
+}
+
+/** definition of our uprobe */
+static int catch_dup(struct uprobe *uprobe, struct upipe *upipe,
+                 int event, va_list args)
```
nto:
```suggestion
                     int event, va_list args)
```

nto (changes requested):
There are still some unaligned function parameters.

nto (approved):

PR #1191, opened 2026-06-24
upipe_disblo: set discontinuity flag after dropping

quarium:


lib/upipe-modules/upipe_discard_blocking.c
```diff
@@ -124,8 +128,12 @@ static void upipe_disblo_input(struct upipe *upipe,
     if (upipe_disblo->nb_urefs >= upipe_disblo->max_urefs) {
         upipe_warn(upipe, "dropping uref");
         uref_free(uref);
+        upipe_disblo->discontinuity = true;
     }
     else {
+        if (upipe_disblo->discontinuity)
+            uref_flow_set_discontinuity(uref);
+        upipe_disblo->discontinuity = false;
```
nto:
```suggestion
        if (upipe_disblo->discontinuity) {
            uref_flow_set_discontinuity(uref);
            upipe_disblo->discontinuity = false;
        }
```

lib/upipe-modules/upipe_discard_blocking.c
```diff
@@ -87,6 +90,7 @@ static struct upipe *upipe_disblo_alloc(struct upipe_mgr *mgr,
 
     struct upipe_disblo *upipe_disblo = upipe_disblo_from_upipe(upipe);
     upipe_disblo->max_urefs = UPIPE_DISBLO_MAX_UREFS_DEFAULT;
+    upipe_disblo->discontinuity = true;
```
nto: Shouldn't that be `false` instead?

nto (changes requested):

lib/upipe-modules/upipe_discard_blocking.c
quarium: Not sure, but imagine you switch a pipe input from an old upipe_disblo to a new one, don't you think you should have a discontinuity?

lib/upipe-modules/upipe_discard_blocking.c
quarium: Why?

nto: For explicitness and efficiency.

lib/upipe-modules/upipe_discard_blocking.c
nto: Switching output does not drop frames, does it?

nto (approved):

PR #1194, opened 2026-07-06
upipe_avfilt: reset sub output latency when reconfiguring

quarium:


lib/upipe-av/upipe_avfilter.c
```diff
@@ -1509,6 +1509,9 @@ static int upipe_avfilt_init_filters(struct upipe *upipe)
             goto end;
         }
 
+        /** reset latency */
```
nto: Comments that only rephrase the code are not useful.

nto (changes requested):

nto (approved):

PR #1204, opened 2026-07-09
Makefile: fix missing deps on static libraries

nto:
Binaries and shared libraries depending on internal static libraries were not relinked properly after said static libraries were rebuilt.


JDarnley:
Thank you.  Works for me for the file I was modifying.

JDarnley (approved):

PR #1214, opened 2026-07-25
uref: allow a uref to carry multiple ubufs

kierank:
Add a chain of additional ubufs to struct uref, linked by their own uchain (unused while a ubuf is in flight), so that a single uref can carry e.g. video, audio and subtitles at the same time.

The primary uref->ubuf is entry 0 and is untouched: legacy pipes keep working unmodified and transparently pass additional entries through, since uref_free frees the chain and uref_dup/uref_fork duplicate it (a cheap reference increment). uref_dup_inner deliberately does not duplicate the chain, so derived urefs (block splice/split, field split) do not fan out the additional payloads.

Per-entry metadata (flow ID, flow def, date, duration) is stored in the parent uref attributes under indexed names ("sub[i].*") and is thus carried and duplicated by the existing udict handling. The new uref_sub.h API provides attach/get/detach/replace/find accessors, with lookup by flow ID.

### DIFF
--- a/include/upipe/uref.h
+++ b/include/upipe/uref.h
@@ -22,6 +22,7 @@ extern "C" {
 #include "upipe/urefcount.h"
 #include "upipe/ubuf.h"
 #include "upipe/udict.h"
+#include "upipe/ulist.h"
 
 #include <assert.h>
 #include <inttypes.h>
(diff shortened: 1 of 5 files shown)


kierank:
Maybe I can dump timestamp and duration as by definition of being in a uref they are the same

funman:
Looks simple enough, I guess use it to see if something is missing

PR #1222, opened 2026-08-04
Dvbsub fixes

kierank:


quarium (approved):

lib/upipe-av/upipe_avcodec_decode.c
```diff
@@ -1095,6 +1095,9 @@ static void upipe_avcdec_output_sub(struct upipe *upipe, AVSubtitle *sub,
             uint8_t *dst = buf + stride * y + 4 * x;
             uint8_t *src = r->data[0];
             uint8_t *palette = r->data[1];
+            /* Source row stride; may exceed the (possibly cropped) blit width,
+             * so advance src by the real stride rather than by w. */
```
nto: The code looks fine but this comment does not make sense. The variable name makes it obvious that it is a stride and comments should not reference previous code that no longer exists.

nto (changes requested):

nto (approved):

PR #1226, opened 2026-08-07
upipe_avfsrc: prevent useless update event

quarium:


nto (changes requested):
`prevent useless update` is vague. It does not explain what update is it about nor why it's useless.

quarium:
Better now?

nto (approved):

PR #1228, opened 2026-08-11
upipe_transfer: do not leak a reference when freeze fails

JDarnley:
Managers allocated without a mutex always hit this: umutex_lock(NULL) returns UBASE_ERR_INVALID, so upipe_work_freeze() bails out through UBASE_RETURN without setting upipe_work->frozen, and upipe_work_thaw() then returns early on !frozen without releasing. The refcount never reaches zero, UPIPE_XFER_DETACH is never sent, and the worker thread never leaves upump_mgr_run(), hanging shutdown.

Lock first and take the reference only once the lock is held. The failing path still returns UBASE_ERR_INVALID and leaves frozen untouched, so inner pipe access is unaffected.


nto:
Duplicate of #1227.

PR #1239, opened 2026-08-25
upipe_avfilt: fix hw support for FFmpeg 8

quarium:


nto (changes requested):
This does not fix anything for avfilter 9.1.100, it's the version where the APIs were made available. The fix is probably for FFmpeg 8.

nto (approved):

PR #1245, opened 2026-08-31
upipe_avcenc: reconfigure on error

quarium:
Some encoder do not fail when opening but when sending the first frame.
It may be because of a temporary resource starvation that require a full
reinitialization.


lib/upipe-av/upipe_avcodec_encode.c
```diff
@@ -180,6 +180,7 @@ struct upipe_avcenc {
     AVPacket *avpkt;
     /** true if the pipe need to be released after output_input */
     bool release_needed;
+    bool flush_needed;
```
nto, line 184: Missing comment.

nto (changes requested):

nto (approved):

